Issue #9, opened 2018-07-24
Should use different brand major for Collection or Sequences


fcartegnie:
Just like other codec mappings:
heic/heix vs hevc/hevx
avci vs avcs

aklemets:
The other file formats have different file extensions for files that include image sequences. It sounds like you are asking for different brand identifiers. But that's not what the other file formats use the brand identifiers for. For example 'heic' and 'heix' is used to distinguish between SDR and HDR, not between sequence and non-sequence files.

lambdapioneer:
According to http://nokiatech.github.io/heif/technical.html (table I), for HEIF files heic/heix indicate an image and hevc/heix indicate a sequence when used in a HEIF file. Being SDR/HDR respectively.

aklemets:
You are right that for HEVC there is a separate brand for image sequences. I had missed that earlier. But I would caution against using the Nokia web site as a reference at the info on that site might be stale. 
I agree with you then that we should add a separate brand for AV1 image sequence files.

cconcolato:
Discussed during 9/26 AOMI meeting. Agreement to define different brands for images vs sequences.

PR #11, opened 2018-08-23
Proposed new version

cconcolato:
This proposed new version is the result of major restructuring and simplification of the spec text, with some technical clarifications.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,17 +1,17 @@
-
 <pre class='metadata'>
 Title: AV1 Image File Format (AVIF)
-Shortname: av1-avif
-Level: 1
-Date: 2018-06-26
-Status: LS-COMMIT
-Group: AOM
+Status: LS
 URL: https://AOMediaCodec.github.io/av1-avif
+Shortname: av1-avif
+Editor: Cyril Concolato, Netflix, cconcolato@netflix.com
+Former Editor: Paul Kerr, Netflix, pkerr@netflix.com
+Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of a [[!MIAF]] profile.
+Date: 2018-06-26
 Repository: AOMediaCodec/av1-avif
 Inline Github Issues: full
-Editor: Paul Kerr, Netflix, pkerr@netflix.com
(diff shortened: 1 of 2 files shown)


cconcolato:
Browsable version here https://rawgit.com/AOMediaCodec/av1-avif/major_restucturing/docs/

index.bs
```diff
@@ -36,564 +36,122 @@ THE OTHER MEMBER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-The image data should have the "reduced_still_picture_header" field set to 1 in the sequence header OBU as specified in section 5.5.1 of the Bitstream & Decoding Process Specification [[!AV1]]. The field may be set to 0 to allow for easy extraction of frames from an AV1 file.
-</p>
-
-<h2 id="avif=image-collection">AVIF Image Collection</h2>
-
-<p>The image data of type "av1i" shall be used for an image collection item coded with AV1.
-
-The image item data shall be structured as defined in the AV1 Sample Format section of the AV1 Codec ISO Media File Format Binding
-[[!AV1]] specification.
+<p>When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and SHALL obey the following constraints:
```
lambdapioneer: Can you share the motivation behind the naming to `av01` compared to `av1i`?

index.bs
```diff
@@ -36,564 +36,122 @@ THE OTHER MEMBER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-</p>
-
-<h3 id="ver1-derived-images">Derived Images</h3>
-
-<p>The derivation of tiled images may be supported by Baseline readers which also support rendering images
-larger than permited by AV1 profiles. No other derived image format need be supported.
-</p>
+<p>An <dfn>AV1 Image Sequence</dfn> is defined as a set of AV1 [=Temporal Units=] stored in an [=AV1 track=] as defined in [[!AV1-ISOBMFF]] with the following constraints:
+<ul>
+<li>The [=Sequence Header OBUs=] SHALL have their <code>[=still_picture=]</code> flag set to 1 and SHOULD have their <code>[=reduced_still_picture_header=]</code> flags set to 1.
```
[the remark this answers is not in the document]
cconcolato: I don't see why we would need timing information inband when the ISOBMFF layer has it. 

Anyway, this text will probably change based on feedback from #12

index.bs
cconcolato: Definitely a discussion we could have. See also #9.

Annex J of HEIF says:
> It is suggested that image sequence tracks are specified like video tracks of the same coding format with a change of handler type (from 'vide' to 'pict') and possibly with additional constraints.

and
>In coded image items, the coding type is identified by the item_type, and in image sequences, by the sample entry type. It is suggested that the same code be used for the two cases;

I just followed these suggestions.

Issue #16, opened 2018-09-13
4:4:4 chroma sub-sampling and profile "High"


lambdapioneer:
**Motivation**

Allowing 4:4:4 Chroma sub-sampling is critical to many application areas (e.g. graphical images or photographic content with overlays). AVIF should allow for storing image information with 4:4:4 chroma sub-sampling.

Comparison with other formats:
 - JPEG: supports 444 (e.g. mozjpeg uses 444 by default for higher quality values)
 - WebP: not supported unless lossless (see https://en.wikipedia.org/wiki/WebP#Restrictions)
 - HEVC: supported with `Main 4:4:4 Intra` profile (see https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding#Version_2_profiles)

**Proposed change**

Change required reader profile to "High" or higher to allow 4:4:4 bit streams.

From the AV1 spec https://aomediacodec.github.io/av1-spec/av1-spec.pdf, p635:

> The Main profile supports YUV 4:2:0 or monochrome bitstreams with bit depth equal to 8 or 10. The Highprofile further adds support for 4:4:4 bitstreams with the same bit depth constraints. Finally, the Professional profile extends support over the High profile to also bitstreams with bit depth equal to 12, and also adds support for the 4:2:2 video format

**Other**

I'd suspect that 444 would play particularly well to the strength of the "chroma from luma prediction" in AV1: https://people.xiph.org/~xiphmont/demo/av1/demo1.shtml

See related issue: #15

cconcolato:
I agree with the general idea. The group needs to have a discussion regarding the number of profiles/brands we want to have and their perimeter.

mzso:
Why not have it as default? The codec would/could do better at throwing away color information than plain dumb 4:2:0/4:2:2 subsampling.

lambdapioneer:
> Why not have it as default? The codec would/could do better at throwing away color information than plain dumb 4:2:0/4:2:2 subsampling.

Setting a default would be part of an encoder implementation. The specification simply says what's allowed :) And I feel that we don't want to exclude 4:2:0 to allow efficient extraction of frames from AV1 video.

If we have some documentation that shows that 4:4:4 out-performs 4:2:0 on AV1, we could add a note or a recommendation.

PR #17, opened 2018-09-13
Allow high profile for 'pict' tracks

lambdapioneer:
Allowing high profile for pict tracks to enable use of 444 chroma sub-sampling. See issue #16 for motivation.

Also removing the "at Main tier" addition to the level constraint, as AV1 profiles and levels are independent of each other.

> The profile specifies the bit depth and subsampling formats supported, while the level defines resolution and performance characteristics. (https://aomediacodec.github.io/av1-spec/av1-spec.pdf, p647)

Not changing for the video tracks for now. The reason is that we'd want to be more conservative there as the decoder needs to playback in real time.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -153,5 +153,5 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in 
 
 <p>If the brand <code>MiA1</code> is in the list of compatible_brands of a TrackTypeBox or FileTypeBox, the following constraints SHALL be respected:
 - The following "shared constraints" of [[!MIAF]] apply: self-containment, single-layer, grid-limit, single-track, and matched-duration.
-- For AV1 images stored as items or in 'pict' tracks, the profile SHALL be the Main Profile and the level SHALL be 6.0 level at Main tier or lower.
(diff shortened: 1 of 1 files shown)


cconcolato:
As indicated in #16 we should discuss the general question of profiles/levels and then implement that in the spec.



> Also removing the "at Main tier" addition to the level constraint, as AV1 profiles and levels are independent of each other.

I don't understand why. "Main tier" refers to the the "Main" **tier** within the level. It is independent of the profile. Maybe the terms "Main" and "High" for tiers and profiles is confusing, but there is nothing we can do about it ...

lambdapioneer:
From my understanding, the "Main tier" and "High tier" only affect the bitrate which doesn't seem to be relevant for still pictures as we don't have real-time requirements.

cconcolato:
> From my understanding, the "Main tier" and "High tier" only affect the bitrate which doesn't seem to be relevant for still pictures as we don't have real-time requirements.

Good point. Maybe worth note.

cconcolato:
Closing as I think your request has been addressed in #21

PR #21, opened 2018-09-20
Addressed various issues, some restructuring

### LINKED ISSUE #9: Should use different brand major for Collection or Sequences

### LINKED ISSUE #12: still_picture =1 should not be a requirement for image sequences
When HEVC is used in a HEIF image sequence, non-IDR frames are allowed. I would expect the same to be allowed with AV1.

### LINKED ISSUE #14: still_picture and reduced_still_picture_header in items
As discussed during today's AOM Image call:
- why do we require still_picture flag set to 1 in item data? That requires flipping this bit when creating an image item from a KEY FRAME in a video sequence. What do we gain by having this flag set?
- why are we recommending to have reduced_still_picture_header flag set to 1? Should leave freedom to encoders to do what they want.

### LINKED ISSUE #19: Distinguish between picture tracks and video tracks
The spec currently treats picture tracks and video tracks as they are the same. I think a picture track can be a sequence of burst images, for example, and would be generated "picture module" of a camera. But a video track is just a plain video track that would be generated by the "video module".
So, while it would be OK to have special restrictions on "picture tracks" (for example, I-frames only), "video tracks" should be mostly unrestricted as they would be created by the camera's normal video module.

### LINKED ISSUE #20: Spec should define file extensions and MIME types
The spec should define file extensions and MIME types. It currently does not.

aklemets:
- Defined Baseline and Advanced profiles for the file format. Moved some requirements from the main section to the profile sections. This partially addresses #16.
- The relationship between this spec and MIAF was a bit unclear. I tried to address this in the definitions of the two profiles.
- Added file extension and MIME type definitions. This closes #20.
- I define a separate brand, file extension and MIME type for image sequences. This closes #9.
- Made the spec separate between pict tracks and video tracks. This closes #19.
- Made still_picture = 1 recommended instead of mandatory. This closes #12 and also closes #14.

I am having some trouble with the Bikeshed syntax. It is currently giving me some warnings and I think I need help clearing that up.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4,9 +4,10 @@ Status: LS
 URL: https://AOMediaCodec.github.io/av1-avif
 Shortname: av1-avif
 Editor: Cyril Concolato, Netflix, cconcolato@netflix.com
+Editor: Anders Klemets, Microsoft, Anders.Klemets@microsoft.com
 Former Editor: Paul Kerr, Netflix, pkerr@netflix.com
 Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image file format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of profiles that extend the Multi-Image Application Format [[!MIAF]].
-Date: 2018-09-19
+Date: 2018-10-11
 Repository: AOMediaCodec/av1-avif
 Inline Github Issues: full
 Boilerplate: property-index no, issues-index no, copyright yes
@@ -110,7 +111,7 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
 
 The [=AV1 Sample=] that is the coded image of the [=AV1 Image Item=], is called <dfn>AV1 Image Item Data</dfn>. 
 The following constraints apply to [=AV1 Image Item Data=]:
-  - The [=AV1 Image Item Data=] shall be signaled as a sync sample, as defined in [[!AV1-ISOBMFF]].
+  - The [=AV1 Image Item Data=] shall be an AV1 sync sample, as defined in [[!AV1-ISOBMFF]].
   - The [=AV1 Image Item Data=] shall have exactly one [=Sequence Header OBU=].
   - The [=AV1 Image Item Data=] should have its <code>[=still_picture=]</code> flag set to 1.
   - The [=AV1 Image Item Data=] should have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
@@ -129,10 +130,10 @@ The following constraints apply to [=AV1 Image Item Data=]:
 
 The syntax and semantics of the <dfn>AV1 Item Configuration Property</dfn> are identical to those of the [=AV1CodecConfigurationBox=] defined in [[!AV1-ISOBMFF]], with the following constraints:
 
-  - [=Sequence Header OBUs=] should not be present in the [=AV1 Item Configuration Property=].
+  - [=Sequence Header OBUs=] should not be present in the [=AV1CodecConfigurationBox=].
+  - If a [=Sequence Header OBU=] is present in the [=AV1CodecConfigurationBox=], it shall match the  [=Sequence Header OBU=] in the [=AV1 Image Item Data=].
   - The values of the fields in the [=AV1CodecConfigurationBox=] shall match those of the [=Sequence Header OBU=] in the [=AV1 Image Item Data=].
   - [=Metadata OBUs=], if present, shall match the values given in other item properties, such as the PixelInformationProperty or ColourInformationBox.
-  - If a [=Sequence Header OBU=] is present in the [=AV1CodecConfigurationBox=], it shall match the  [=Sequence Header OBU=] in the [=AV1 Image Item Data=].
 
 This property shall be marked as essential.
 
@@ -153,21 +154,21 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in 
 
     <h2 id="alpha-images">Alpha Image Items and Sequences</h2>
 <p>An <dfn>AV1 Alpha Image Item</dfn> (respectively an <dfn>AV1 Alpha Image Sequence</dfn>) is an [=AV1 Image Item=] (respectively [=AV1 Image Sequence=]</dfn>) with the following additional constraints:
-    - The <code>[=mono_chrome=]</code> field in the [=Sequence Header OBU=] SHALL be set to 1.
+    - The <code>[=mono_chrome=]</code> field in the [=Sequence Header OBU=] shall be set to 1.
 
-<p>In [[!ISOBMFF]] or [[!HEIF]] files, if an [=AV1 Alpha Image Item=] (respectively. an [=AV1 Alpha Image Sequence=]) is used, the <code>aux_type</code> field of the <code>AuxiliaryTypeProperty</code> (respectively <code>AuxiliaryTypeInfoBox</code>) SHALL be set to <code>urn:mpeg:mpegB:cicp:systems:auxiliary:alpha</code>, as defined in [[!MIAF]].</p>
+<p>In [[!ISOBMFF]] or [[!HEIF]] files, if an [=AV1 Alpha Image Item=] (respectively. an [=AV1 Alpha Image Sequence=]) is used, the <code>aux_type</code> field of the <code>AuxiliaryTypeProperty</code> (respectively <code>AuxiliaryTypeInfoBox</code>) shall be set to <code>urn:mpeg:mpegB:cicp:systems:auxiliary:alpha</code>, as defined in [[!MIAF]].</p>
 
   <h2>Brands and file extensions</h2>
 
 <h3>AVIF image and image collection brand</h3>
-Files conformant with the brand-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) shall include the brand
+Files that conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) should include the brand
 <dfn value="" export="" for="AVIF Image brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
 <p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
 <p>If 'avif' is specified in the major_brand field of the FileTypeBox, the file extension should be ".avif".
 The MIME Content Type for files with the ".avif" file extension shall be "image/avif".</p>
 
 <h3>AVIF image sequence brand</h3>
-Files conformant with the brand-independent restrictions in this document (sections [[#image-item-and-properties]], [[#image-sequences]] and [[#alpha-images]]) shall include the brand
+Files which contain one or more image sequences, and which conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]], [[#image-sequences]] and [[#alpha-images]]), should include the brand
 <dfn value="" export="" for="AVIF Image Sequence brand">avis</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
 <p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
 <p>If 'avis' is specified in the major_brand field of the FileTypeBox, the file extension should be ".avifs".
--- a/index.bs
+++ b/index.bs
@@ -155,15 +155,15 @@ Files conformant with the brand-independent restrictions in this document shall 
 
   <h3 id="image-and-image-collection-files">AV1 image and image collection files</h3>
   The file extension ".avif" and the MIME Content Type "image/avif" should be used for AVIF files that satisfy the following requirements:
-  - The file does not contain image sequences of any kind.
-  - The [=primary image=] is of type 'av01'.
+  - The file does not contain tracks, including image sequences, of any kind.
+  - The [=primary image=] is either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
   - The AVIF file conforms to at least one AVIF profile, and the profile is listed in the [=compatible_brands=] field of the [=FileTypeBox=].
   - The AVIF file specifies 'avif' in the [=major_brand=] field of the [=FileTypeBox=].
 
   <h3 id="image-sequence-files">AV1 image sequence files</h3>
   The file extension ".avifs" and the MIME Content Type "image/avif-sequence" should be used for AVIF files that satisfy the following requirements:
   - The file contains at least one [=AV1 Image Sequence=].
-  - The [=primary image=] is of type 'av01'.
+  - The [=primary image=] is either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
   - The AVIF file conforms to at least one AVIF profile, and the profile is listed in the [=compatible_brands=] field of the [=FileTypeBox=].
   - The AVIF file specifies 'avif' in the [=major_brand=] field of the [=FileTypeBox=].
 
@@ -179,7 +179,7 @@ AVIF files that do not conform to at least one AVIF profile shall not use the ".
 
   The following constraints are common to all profiles defined in this specification:
   - The file shall be compliant with the [[!MIAF]] specification and list 'miaf' in the [=compatible_brands=] field of the [=FileTypeBox=].
-  - The [=primary image=] shall be an [=AV1 Image Item=].
+  - The [=primary image=] shall be either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
 
   <h3 id="AVIF-baseline-profile">AVIF Baseline Profile</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -5,8 +5,8 @@ URL: https://AOMediaCodec.github.io/av1-avif
 Shortname: av1-avif
 Editor: Cyril Concolato, Netflix, cconcolato@netflix.com
 Former Editor: Paul Kerr, Netflix, pkerr@netflix.com
-Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of a [[!MIAF]] profile.
-Date: 2018-08-29
+Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image file format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of profiles for the Multi-Image Application Format [[!MIAF]].
+Date: 2018-09-19
 Repository: AOMediaCodec/av1-avif
 Inline Github Issues: full
 Boilerplate: property-index no, issues-index no, copyright yes
@@ -15,7 +15,7 @@ Group: AOM
 Warning: Custom
 Custom Warning Title: Warning
 Custom Warning Text: This specification is still at draft stage and should not be referenced other than as a working draft.
-Status Text: Version A.1
+Status Text: Version A.2
 </pre>
 
 <div boilerplate='copyright'>
@@ -36,14 +36,23 @@ THE OTHER MEMBER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 </div>
 
 <pre class="anchors">
-url: http://iso.org/#; spec: MIAF; type: property;
-    text: clli
+url: http://iso.org/#; spec: HEIF; type: property;
     text: colr
-    text: mdcv
-    text: miaf
+    text: image item
     text: pasp
     text: pict
     text: pixi
+
+url: http://iso.org/#; spec: ISOBMFF; type: property;
+    text: compatible_brands
+    text: FileTypeBox
+    text: major_brand
+
+url: http://iso.org/#; spec: MIAF; type: property;
+    text: clli
+    text: mdcv
+    text: miaf
+    text: primary image
 
 url: https://aomediacodec.github.io/av1-isobmff/#; spec: AV1-ISOBMFF; type: dfn;
     text: av1codecconfigurationbox
@@ -64,9 +73,9 @@ url: http://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn;
 
 <h2 id="general">Scope</h2>
 
-[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn>AV1 Image File Format</dfn> defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=]. Specifically, only bitstreams with the <code>[=still_picture=]</code> flag set to 1 are supported. For some other configurations of AV1 bitstreams, the [[!AV1-ISOBMFF]] specification SHOULD be used.
-
-ISSUE: Is the restriction to have still_picture = 1 necessary?
+[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn>AV1 Image File Format</dfn> (AVIF) defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=] in a [[!HEIF]] file.
+The [=AV1 Image File Format=] defines multiple profiles, which restrict the allowed syntax and semantics of the [=AV1 bitstream=].
+The profiles defined in this specification follows the convention of the [[!MIAF]] specification.
 
 [=AV1 Image File Format=] supports High Dynamic Range (HDR) and Wide Color Gamut (WCG) images as well as Standard Dynamic Range (SDR). It supports monochrome images as well as multi-channel images with all the bit depths and color spaces specified in [[!AV1]].
 
@@ -74,8 +83,6 @@ An AVIF file is designed to be a conformant [[!HEIF]] file for both image items 
 New Image Formats and Brands" of [[!HEIF]].
 
 This specification reuses syntax and semantics used in [[!AV1-ISOBMFF]].
-
-The MIAF AV1 profile defined in this specification follows the convention of the [[!MIAF]] specification.
 
 ISSUE: Should this specification also define a structural brand stricter than 'miaf' to profile out some features?
 
@@ -94,12 +101,7 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
 
 <h3 id="image-item">AV1 Image Item</h3>
 
-<p>When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and SHALL obey the following constraints:
-<ul>
-<li>The item data SHALL be the content of a single valid [=AV1 Sync Sample=] as defined in [[!AV1-ISOBMFF]] with the additional constraints that there SHALL be only one [=Sequence Header OBU=] and it SHALL have its <code>[=still_picture=]</code> flag set to 1 and SHOULD have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
-<li>The item SHALL be associated with an [=AV1 Item Configuration Property=].</li>
-</ul>
-
+<p>When an image item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and shall be associated with an [=AV1 Item Configuration Property=].
 </p>
 
 <h3 id="image-item-properties">Image Item Properties</h3>
@@ -117,7 +119,7 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
 The syntax and semantics of the <dfn>AV1 Item Configuration Property</dfn> are identical to those of the [=AV1CodecConfigurationBox=] defined in [[!AV1-ISOBMFF]], with the following constraints:
 <ul>
 <li>[=Sequence Header OBUs=] SHOULD not be present in the [=AV1 Item Configuration Property=].</li>
-<li>The values of the fields in the [=AV1CodecConfigurationBox=] SHALL match those of the [=Sequence Header OBU=] in the item data and in the [=AV1CodecConfigurationBox=], if any.</li>
+<li>The values of the fields in the [=AV1CodecConfigurationBox=] SHALL match those of the [=Sequence Header OBU=], if any.</li>
 <li>[=Metadata OBUs=], if present, SHALL match the values given in other item properties, such as the PixelInformationProperty or ColourInformationBox.</li>
 </ul>
 This property SHALL be marked as essential.
@@ -131,12 +133,8 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in 
 
 <h2 id="image-sequences">Image Sequences</h2>
 
-<p>An <dfn>AV1 Image Sequence</dfn> is defined as a set of AV1 [=Temporal Units=] stored in an [=AV1 track=] as defined in [[!AV1-ISOBMFF]] with the following constraints:
-<ul>
-<li>The [=Sequence Header OBUs=] SHALL have their <code>[=still_picture=]</code> flag set to 1 and SHOULD have their <code>[=reduced_still_picture_header=]</code> flags set to 1.
-</li>
-<li>The track handlers <code>vide</code> or <code>pict</code> MAY be used.
-</ul>
+<p>An <dfn>AV1 Image Sequence</dfn> is defined as a set of AV1 [=Temporal Units=] stored in an [=AV1 track=] as defined in [[!AV1-ISOBMFF]].
+  The track handler for an [=AV1 Image Sequence=] shall be <code>pict</code>.
 
 <h2 id="alpha-images">Alpha Image Items and Sequences</h2>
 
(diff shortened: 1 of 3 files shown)


cconcolato:
Online version here for review https://rawgit.com/AOMediaCodec/av1-avif/aklemets-update1/

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  <h3 id="image-sequence-files">AV1 image sequence files</h3>
+  The file extension ".avifs" and the MIME Content Type "image/avif-sequence" should be used for AVIF files that satisfy the following requirements:
+  - The file contains at least one [=AV1 Image Sequence=].
+  - The [=primary image=] is either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
+  - The AVIF file conforms to at least one AVIF profile, and the profile is listed in the [=compatible_brands=] field of the [=FileTypeBox=].
+  - The AVIF file specifies 'avif' in the [=major_brand=] field of the [=FileTypeBox=].
+
+  <h3 id="non-conforming-files">AVIF files not confirming to a profile</h3>
+AVIF files that do not conform to at least one AVIF profile shall not use the ".avif" or ".avifs" file extensions. Such files may use the ".heif" or ".heifs" file extensions, as defined in Annex D and E of [[!HEIF]].
```
lambdapioneer: Might have an non-up-to-date version, but I think it's Annex C and D

index.bs
```diff
@@ -64,9 +73,9 @@ url: http://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn;
 
 <h2 id="general">Scope</h2>
 
-[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn>AV1 Image File Format</dfn> defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=]. Specifically, only bitstreams with the <code>[=still_picture=]</code> flag set to 1 are supported. For some other configurations of AV1 bitstreams, the [[!AV1-ISOBMFF]] specification SHOULD be used.
-
-ISSUE: Is the restriction to have still_picture = 1 necessary?
+[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn>AV1 Image File Format</dfn> (AVIF) defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=] in a [[!HEIF]] file.
```
lambdapioneer: `in a [[!MIAF]] file.`?

index.bs
aklemets: I am using 2nd Ed. of the HEIF spec (ISO/IEC 23008-12) and the MIME type registration is in Annex D and E.

index.bs
aklemets: This is a bit confusing, but I believe the AVIF spec is supposed define a basic mapping of AV1 to HEIF and then also define Profiles that require adherence to the MIAF spec. The MIAF spec in some sense is not a different file format but just a set of restrictions on HEIF.

It is possible, that someone may find a need to put AV1 in a HEIF container but use extra features that are not compatible with MIAF. In that case, they could still use the general provisions of the AVIF spec, but they would not implement any of the Profiles.

In other words, the idea is to make the profile-independent section of the AVIF spec reference HEIF instead of MIAF to allow for future flexibility.

index.bs
aklemets: Regarding the comments about profiles, I am not familiar with the history of WebP, but "profiles" are common in standards. See for instance HEIF, MIAF, H.265, etc. There are plenty of profiles there. The idea for AVIF specifically is to put the stuff that we think all implementations reasonably could support in the Base profile, and then have an Advanced profile that includes features that more than one implementation wants to support but which are unlikely supported by all implementations.

My assumption -- perhaps mistaken -- is that everyone will support Main profile but High profile might not be universally supported. (If everyone can support High profile then why does AV1 define two separate profiles?)

Now, it could be that I am wrong and everyone will support AV1 High profile, in which case we could put AV1 High profile in the AVIF Baseline profile, and AV1 Professional profile in the AVIF Advanced profile.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  <h2 id="brands-and-file-extensions">Brands and file extensions</h2>
+
+  <h3 id="AVIF-general-brand">AVIF General Brand</h3>
+
+Files conformant with the brand-independent restrictions in this document shall include the brand
+<dfn value="" export="" for="AVIF General brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
+  Files shall also carry a compatible brand to identify the AVIF profile, if any, with which the file complies.
+
+  <h3 id="image-and-image-collection-files">AV1 image and image collection files</h3>
+  The file extension ".avif" and the MIME Content Type "image/avif" should be used for AVIF files that satisfy the following requirements:
```
cconcolato: Do we really want a separate MIME type? Can't we use HEIF with codecs = av1?

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The file does not contain tracks, including image sequences, of any kind.
+  - The [=primary image=] is either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
+  - The AVIF file conforms to at least one AVIF profile, and the profile is listed in the [=compatible_brands=] field of the [=FileTypeBox=].
```
cconcolato: Why? If the "Professional" profile is not formally defined in AVIF but one wants to create an AVIF file with a "Professional" image in it, they would not be allowed to use this general brand. That does not seem right.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  <h3 id="AVIF-general-brand">AVIF General Brand</h3>
+
+Files conformant with the brand-independent restrictions in this document shall include the brand
+<dfn value="" export="" for="AVIF General brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
+  Files shall also carry a compatible brand to identify the AVIF profile, if any, with which the file complies.
+
+  <h3 id="image-and-image-collection-files">AV1 image and image collection files</h3>
+  The file extension ".avif" and the MIME Content Type "image/avif" should be used for AVIF files that satisfy the following requirements:
+  - The file does not contain tracks, including image sequences, of any kind.
```
cconcolato: In ISOBMFF/HEIF/MIAF philosophy, we don't forbid other use of a file. I would remove the sentence.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The [=primary image=] is either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
+  - The AVIF file conforms to at least one AVIF profile, and the profile is listed in the [=compatible_brands=] field of the [=FileTypeBox=].
+  - The AVIF file specifies 'avif' in the [=major_brand=] field of the [=FileTypeBox=].
```
cconcolato: Why? A file can be conformant to the brand without having it as a major brand.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  <h3 id="image-sequence-files">AV1 image sequence files</h3>
+  The file extension ".avifs" and the MIME Content Type "image/avif-sequence" should be used for AVIF files that satisfy the following requirements:
+  - The file contains at least one [=AV1 Image Sequence=].
+  - The [=primary image=] is either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
```
cconcolato: Why constrain the primary item in an image sequence file?

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The AVIF file conforms to at least one AVIF profile, and the profile is listed in the [=compatible_brands=] field of the [=FileTypeBox=].
```
cconcolato: same remark as for the image file. What about image sequences using AV1 but not conforming to any profile defined today in AVIF?

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The AVIF file specifies 'avif' in the [=major_brand=] field of the [=FileTypeBox=].
```
cconcolato: Same as above. Why mandate it as a major brand?

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  <h3 id="non-conforming-files">AVIF files not confirming to a profile</h3>
```
cconcolato: typo: conforming

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  <h2 id="profiles-and-brands">Profiles and brands</h2>
+
+  The profiles defined in this section are for enabling interoperability between [=AV1 Image File Format=] files and [=AV1 Image File Format=] readers/parsers.
+  A profile imposes a set of specific restrictions.
+  The presence of a brand indicating a profile can be interpreted as the permission for those [=AV1 Image File Format=] readers/parsers and [=AV1 Image File Format=] renderers that only implement the features required by the profile, to process the corresponding file.
+  An [=AV1 Image File Format=] file may conform to multiple profiles. Similarly, an [=AV1 Image File Format=] reader/parser or [=AV1 Image File Format=] renderer may be capable of processing one or more profiles.
+
+  The following constraints are common to all profiles defined in this specification:
+  - The file shall be compliant with the [[!MIAF]] specification and list 'miaf' in the [=compatible_brands=] field of the [=FileTypeBox=].
```
cconcolato: Compliance to MIAF should be declared by the use of MIAF brands.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The [=primary image=] shall be either an [=AV1 Image Item=] or a derived image that references one or more coded images that all are [=AV1 Image Items=].
+
+  <h3 id="AVIF-baseline-profile">AVIF Baseline Profile</h3>
+
+<p>This section defines the MIAF AV1 Baseline profile of [[!HEIF]], specifically for [[!AV1]] bitstreams, based on the constraints specified in [[!MIAF]] and identified by the brand  <dfn value export for="AV1 Image Item Type">MA1B</dfn>.</p>
+
+<p>If the brand <code>MA1B</code> is in the list of [=compatible_brands=] of a TrackTypeBox or [=FileTypeBox=], the common constraints in the Profiles and brands section shall apply.
+
+  The following additional constraints apply to all [=AV1 Image Items=] and all [=AV1 Image Sequences=]:
+  - The [=AV1 Image Item=] is the content of a single valid [=AV1 Sync Sample=] as defined in [[!AV1-ISOBMFF]].
```
cconcolato: See comment above. It should be in the general section defining AV1 Image Items. Why do you think it should be brand specific? Also it is duplicated in the profiles in this PR.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The [=AV1 Image Item=] shall have only one [=Sequence Header OBU=].
```
cconcolato: Ditto

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The [=AV1 Image Item=] shall have its <code>[=still_picture=]</code> flag set to 1.
```
cconcolato: ditto, plus as agreed this morning, 'shall' should be changed to 'should'

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The [=AV1 Image Item=] should have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
```
cconcolato: ditto

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The AV1 profile shall be the Main Profile and the level shall be 5.1 level at Main tier or lower.
```
cconcolato: As discussed, we should add a NOTE saying that tiers are not constrained because not relevant, given that timing is optional in image sequence and not relevant in image items/collections.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  ISSUE: Is the restriction to have still_picture = 1 necessary?
```
cconcolato: This can be removed.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+  The following additional constraints apply only to AV1 images stored in a 'vide' track:
```
cconcolato: As discussed this morning. We do not constrain in AVIF the content of 'vide' tracks. This paragraph should be removed.

index.bs
```diff
@@ -147,11 +145,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+  - The AV1 profile level shall be Main Profile and the level shall be 5.1 level at Main tier or lower.</p>
+
+  ISSUE: Review profile and level restrictions for all profiles. Level 5.1 is chosen for the Baseline profile to ensure that no single coded image exceeds 4k resolution, as some decoder may not be able to handle larger images. It is still possible to use Baseline profile to create larger images using grid derivation.
```
cconcolato: The issue should be removed and the second sentence and the ones after should be turned into a NOTE.

index.bs
```diff
@@ -117,7 +119,7 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
 The syntax and semantics of the <dfn>AV1 Item Configuration Property</dfn> are identical to those of the [=AV1CodecConfigurationBox=] defined in [[!AV1-ISOBMFF]], with the following constraints:
 <ul>
 <li>[=Sequence Header OBUs=] SHOULD not be present in the [=AV1 Item Configuration Property=].</li>
-<li>The values of the fields in the [=AV1CodecConfigurationBox=] SHALL match those of the [=Sequence Header OBU=] in the item data and in the [=AV1CodecConfigurationBox=], if any.</li>
```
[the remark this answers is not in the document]
aklemets: There was some grammar error in the sentence, and I couldn't make sense of it, so I cleaned it up. The original sentence basically said that that "the fields in A shall match those of B and in A". Of course A matches A, so that is redundant and was removed.

index.bs
```diff
@@ -131,12 +133,8 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
 
 <h2 id="image-sequences">Image Sequences</h2>
 
-<p>An <dfn>AV1 Image Sequence</dfn> is defined as a set of AV1 [=Temporal Units=] stored in an [=AV1 track=] as defined in [[!AV1-ISOBMFF]] with the following constraints:
-<ul>
-<li>The [=Sequence Header OBUs=] SHALL have their <code>[=still_picture=]</code> flag set to 1 and SHOULD have their <code>[=reduced_still_picture_header=]</code> flags set to 1.
```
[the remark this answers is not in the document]
aklemets: Agreed, now that "still_picture" = 1 will say "should" instead of "shall".

index.bs
aklemets: The "codecs" parameter isn't universally recognized and isn't even explicitly mentioned as something that may occur on the image/heif MIME type, so, yes, I think it will be very helpful to have a separate MIME type and file extension. (Basically, I have seen the "codecs" parameter used in MPEG-DASH but it is not used much outside of that feature area.)

index.bs
aklemets: That's possible but that means it would be allowed for a file that uses the still-image brand to include an image sequence. Clients might just ignore the image sequence in that case.

index.bs
aklemets: This text is appearing in the section that defines the rules for the ".avif" file extension. It says that if you use the ".avif" file extension, the file needs to have a Primary Image that matches one of the profiles in this spec. It does not prohibit the use AV1 Professional Profile in a file that uses the 'avif' brand. But such a file shouldn't use the ".avif" file extension. It can use the more general ".heif" or ".hif" file extensions instead.

index.bs
aklemets: Yes, you are right that it can. However, this text is appearing in the section that defines the rules for the ".avif" file extension and since this is a new file extension we have the opportunity to button down the rules a little bit and require that 'avif' be specified if as the major_brand if you want to use the .avif file extension. The problem with the ISOBMFF file type box is that it has variable size (can contain up to 32 different brands, I think) and this makes it difficult for implementations that try to recognize file types by looking for a well-known byte pattern at a well-known offset in the file.
There is a risk that if most, but not all, implementations put 'avif' in the major_brand box, that some clients will expect that 'avif' will always appear in the major_brand box. Those implementations might break if someone creates a .avif file with 'avif' only in the compatible_brands box. This potential interop problem is avoided by defining stricter rules right now, and require that 'avif' be in the major_brand box.

index.bs
aklemets: Yes, that is what the section is saying. The 'miaf' brand is used to declare compliance with MIAF.

index.bs
aklemets: I moved this to the profile-specific section because the restriction that says that AV1 images must be sync samples is too restrictive for the image tracks. The idea is that we may want to have other profiles in the future were all images in an image track don't need to be sync samples.
However, at the recent conference call there seemed to be consensus that image tracks will allow the same types of images as video tracks, e.g., predicted images will now be allowed. So, in that case I agree that it does not need to be profile-specific anymore. I will move it back.

index.bs
aklemets: Yes

index.bs
aklemets: Agreed, can move it back now

index.bs
aklemets: Yes

index.bs
aklemets: My mistake, I meant to say 'pict' track here.

index.bs
aklemets: Actually, this text, which appears in section 5.3, also appears in section 6. But the next bullet that says that the file must conform to one of the profiles, implicitly references section 6, so to avoid duplication I should remove this text from 5.3.

index.bs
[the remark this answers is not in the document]
aklemets: I think I understand now what the original sentence was trying to say. I have added a separate bullet item to make it clearer.

aklemets:
Cyril, I replied to each of your comments and made changes where appropriate, and pushed a new changes. Please review the updated version.

index.bs
```diff
@@ -5,8 +5,8 @@ URL: https://AOMediaCodec.github.io/av1-avif
 Shortname: av1-avif
 Editor: Cyril Concolato, Netflix, cconcolato@netflix.com
 Former Editor: Paul Kerr, Netflix, pkerr@netflix.com
-Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of a [[!MIAF]] profile.
-Date: 2018-08-29
+Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image file format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of profiles for the Multi-Image Application Format [[!MIAF]].
```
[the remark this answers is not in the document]
cconcolato: I don't understand what you mean by "for" in "they are actual profiles for the MIAF format".  The MIAF spec says "External organizations may define additional MIAF based profiles". That is what our spec is doing. "extend" can be confusing as it could mean that the profiles defined in AVIF extend the profiles defined in MIAF. What about:
> profiles based on constraints defined in the Multi-Image Application Format.

index.bs
```diff
@@ -64,9 +73,9 @@ url: http://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn;
 
 <h2 id="general">Scope</h2>
 
-[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn>AV1 Image File Format</dfn> defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=]. Specifically, only bitstreams with the <code>[=still_picture=]</code> flag set to 1 are supported. For some other configurations of AV1 bitstreams, the [[!AV1-ISOBMFF]] specification SHOULD be used.
-
-ISSUE: Is the restriction to have still_picture = 1 necessary?
+[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn>AV1 Image File Format</dfn> (AVIF) defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=] in a [[!HEIF]] file.
+The [=AV1 Image File Format=] defines multiple profiles, which restrict the allowed syntax and semantics of the [=AV1 bitstream=].
+The profiles defined in this specification follows the convention of the [[!MIAF]] specification.
```
cconcolato: typo: "follow the conventions"

index.bs
```diff
@@ -94,13 +101,13 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
 <h3 id="image-item">AV1 Image Item</h3>
 
-<p>When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and SHALL obey the following constraints:
-<ul>
-<li>The item data SHALL be the content of a single valid [=AV1 Sync Sample=] as defined in [[!AV1-ISOBMFF]] with the additional constraints that there SHALL be only one [=Sequence Header OBU=] and it SHALL have its <code>[=still_picture=]</code> flag set to 1 and SHOULD have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
-<li>The item SHALL be associated with an [=AV1 Item Configuration Property=].</li>
-</ul>
-
-</p>
+When an image item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and shall obey the following constraints:
```
cconcolato: I tried to be consistent with the AV1-ISOBMFF spec where we use all capitals for normative words: SHOULD, SHALL,... This is also the way they are spelled in the conformance section https://aomediacodec.github.io/av1-avif/#conformance. Is this a problem?

index.bs
```diff
@@ -94,13 +101,13 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
+    
+  - The [=AV1 Image Item=] is the content of a single valid [=AV1 Sync Sample=] as defined in [[!AV1-ISOBMFF]].
```
cconcolato: There is a difference between the item (which is a logical entity) vs. the item data which is the physical one. The use of item data here was meant to emphasize that the data has to be physically formed by the content of sync sample.

index.bs
```diff
@@ -94,13 +101,13 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
+  - The [=AV1 Image Item=] shall have exactly one [=Sequence Header OBU=].
+  - The [=AV1 Image Item=] should have its <code>[=still_picture=]</code> flag set to 1.
+  - The [=AV1 Image Item=] should have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
+  - The item shall be associated with an [=AV1 Item Configuration Property=].
```
cconcolato: Here you can use [=AV1 Image Item=]

index.bs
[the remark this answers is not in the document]
cconcolato: Sorry the last A was meant to be configOBUs. We may have the same information at 3 places: 
- in the Configuration record as direct fields
- in the OBUs in the configOBUs field of the Configuration record
- in the OBUs in the item data
They all shall match.

index.bs
aklemets: Yes, the conformance section is says that these words are supposed to be in lower case. I am quoting, "However, for readability, these words do not appear in all uppercase letters in this specification.". So, I am just trying to be consistent with what the spec already says.

index.bs
cconcolato: > The "codecs" parameter isn't universally recognized

I think it is for ISOBMFF-based files.

> isn't even explicitly mentioned as something that may occur on the image/heif MIME type

Annex D and Annex E both define the 'codecs' parameter. It is also in the format IANA registry: https://www.iana.org/assignments/media-types/image/heif
Note that there are 3 other sub-parameters: profiles, itemtypes and lhevcptl. The first two could make sense for AVIF too.

index.bs
aklemets: I did understand that you had a comment on that (maybe it was a different comment?) The idea is that if you have a file that uses the file extension for still image files, there should not be image tracks in that file, because there is a separate file extension for that. The idea was to have a rule to force people to use the correct file extension for image-track files.
But I agree that having a restrictive rule like this does not follow established pattern, so I have already removed this sentence from the second iteration that you are reviewing now.

index.bs
cconcolato: > can contain up to 32 different brands, I think

No. It's unlimited.

I'm a bit uneasy as this goes beyond the practice in the file format and HEIF/MIAF. 

For example, HEIF says the following about the 'mif1'/'msf1' brands (A.2.1.1/A.3.1.1):
> When this brand is the major brand, the defined file extension and MIME type should be used.

Also all MIME type definitions (e.g. Annex D) indicate:
> The file extension and MIME type of a file in this family usually reflect the major brand in the FileTypeBox.

Also, MIAF says:
> A MIAF file shall use the filename extensions specified by HEIF to identify the presence of specific image coding formats, as summarized in Table 2.

This does not seem to leave room for defining new extensions. Maybe @dwsinger can comment here.

index.bs
aklemets: When I said the codecs parameter is not universally recognized, it was not a conjecture. I am responsible for media file MIME types in Windows and on Xbox One, so I should know... If we want to design a highly interoperable file format, then we need to have a separate file extension for this file format.

index.bs
cconcolato: That's what I find strange. If we define a file extension, it should be for all AV1 content. Then you look at the filetypebox (or codecs parameter) and see if you find a profile that you support.

index.bs
aklemets: It does leave room for defining new extensions. They simply go in the compatible_brands field, as usual. The quote from the MIME type definitions support the idea that the file extension should reflect the major brand. I.e., if the file extension is ".avif" then the major brand is supposed to be 'avif', and vice versa.

index.bs
cconcolato: ISOBMFF has the concept of dual-headed files, meaning a file may serve several purposes and the compatible brands give you all these purposes. The major brand gives you one of them but tools may rewrite them as long as they keep the compatible brands that won't change the use of the file. I fear that forcing the major brand will restrict what you can do.

index.bs
aklemets: But the file extension isn't always tied to the codec. For example, if I put H264 in a ISOBMFF file in some way that is incompatible with the MP4 spec, I can't use the .mp4 file extension.
Or as another example, if a HEIF file uses JPEG for the Primary Item, and uses AV1 for another item, and HEVC for another item, I can't use the .heic file extension just because there happens to be some HEVC in the file.

index.bs
aklemets: If the tool changes the major brand to something else, it should change the file extension as well to match whatever the new major brand is. I don't think it is a problem. Especially if we are clear about the rule in this spec.

index.bs
cconcolato: There are at least 3 different terms:
- "image" as defined in HEIF, i.e. arrays of pixels. This is a logical term used for rendering purposes
- "image item", i.e. a logical term used for storage purposes.  "image item" and "image" are very similar. I think an image item may carry multiple images in the case of scalability.
- 'item data", i.e. the actual bytes stored in the 'idat' or 'mdat' box. 
I prefer to use "item data" in this sentence.

index.bs
cconcolato: I'm not opposed to defining a file extension. I'm just saying it won't be the only way of delivering AVIF files.

index.bs
cconcolato: > It does leave room for defining new extensions.

How? I was talking about "file extension". Table 2 says for any other codec than AVC or HEVC you must use ".heif" or ".heifs". Although it's confusing as the title of table says informative ...



> The quote from the MIME type definitions support the idea that the file extension should reflect the major brand. I.e., if the file extension is ".avif" then the major brand is supposed to be 'avif', and vice versa.

Yes, I'm beginning to think so.

dwsinger: MIAF is specifically about those conformance points. We might need to ask MPEG to make it clear that external codecs can be structurally-MIAF-compatible, and what compatible brand(s) should be used.

Please don't ever 'claim' the major brand; you can say "if the major brand is X, the extension shall be Y" and so on.  (I think you know this).

index.bs
```diff
@@ -131,27 +138,80 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
 </ul>
 
 <p>In [[!ISOBMFF]] or [[!HEIF]] files, if an [=AV1 Alpha Image Item=] (respectively. an [=AV1 Alpha Image Sequence=]) is used, the <code>aux_type</code> field of the <code>AuxiliaryTypeProperty</code> (respectively <code>AuxiliaryTypeInfoBox</code>) SHALL be set to <code>urn:mpeg:mpegB:cicp:systems:auxiliary:alpha</code>, as defined in [[!MIAF]].</p>
 
-<h2 id="AVIF-baseline-profile">MIAF AV1 Baseline Profile</h2>
+  <h2 id="brands-and-file-extensions">Brands and file extensions</h2>
+
+  <h3 id="AVIF-general-brand">AVIF General Brand</h3>
+
+Files conformant with the brand-independent restrictions in this document shall include the brand
```
cconcolato: We should be clearer about what the "brand-independent restrictions" are. Also, I think it should be phrased the other way around: if a file contains the 'avif' brand in the compatible_brands field, it shall be conformant to section 2, 3 or 4.

cconcolato:
@aklemets could you update the PR with the latest agreements to see where we stand?

aklemets:
Of course, I will look over your feedback and post an update in the next 24 hours or so.

index.bs
aklemets: I understand the distinction, but the problem is that "item data" is not a defined term in any of the specs that this spec references. And introducing a new concept of "item data" in this spec would seem odd. I see that for HEVC, the HEIF spec defines a struct called "HEVCItemData", which is exactly that we would need for AV1. 

Looking at the AV1-ISOBMFF spec, the nearest equivalent to "HEVCItemData" is "AV1 Sample". The definition about "AV1 Sample" is a bit muddled up because it gives the defintion of the item data and talks about Track header all in the same section. But I think it works for our purposes.

The AVIF spec currently does not use "AV1 Sample" but rather uses "AV1 Sync Sample", which is not a defined term in the AV1-ISOBMFF spec.

I will change this section (2.1) to use "AV1 Sample" and you can review again and let me know if looks OK.

index.bs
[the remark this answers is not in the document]
aklemets: Agreed, and I believe the update I made last week already covers what you say. But I think you want to distinguish between the "image item" and the data associated with that item. I think that is the only real difference between what I wrote and what you are asking for.
Since "image data" is not a term that defined anywhere, I am defining an "AV1 Image Item Data" term and using that in the new update. Take a look and let me know what you think...

index.bs
aklemets: Fixing in my next update to this PR.

index.bs
aklemets: Yes, fixed

index.bs
aklemets: I am OK with phrasing it the other way around, although I don't see what the big difference would be. One reason for the current structure is that I had only a single brand for images and images sequences. If I am swapping the phrasing around, it makes more sense if we have seprate brands for images and image sequences. 
Since we had already agreed to that we will have separate brands, I will just go ahead and defined the 'avis' brand for images sequences, and swap the phrasing around as you suggested.

aklemets:
@cconcolato I am done with the latest round of updates. Please take a look again.

index.bs
dwsinger: I support that the technology specification (how to do AV1 in this family) is AV1-in-HEIF, and that the profile restrictions are to match MIAF.

index.bs
```diff
@@ -131,27 +144,72 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+
+<h3>AVIF image and image collection brand</h3>
+Files conformant with the brand-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) shall include the brand
+<dfn value="" export="" for="AVIF Image brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
+<p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
+<p>If 'avif' is specified in the major_brand field of the FileTypeBox, the file extension should be ".avif".
+The MIME Content Type for files with the ".avif" file extension shall be "image/avif".</p>
+
+<h3>AVIF image sequence brand</h3>
+Files conformant with the brand-independent restrictions in this document (sections [[#image-item-and-properties]], [[#image-sequences]] and [[#alpha-images]]) shall include the brand
```
cconcolato: Same comment as above for avif.

index.bs
```diff
@@ -131,27 +144,72 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+<dfn value="" export="" for="AVIF Image Sequence brand">avis</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
+<p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
+<p>If 'avis' is specified in the major_brand field of the FileTypeBox, the file extension should be ".avifs".
+The MIME Content Type for files with the ".avifs" file extension shall be "image/avif-sequence".</p>
```
cconcolato: Same as above, should have an additional section for definition of the MIME type

index.bs
```diff
@@ -5,8 +5,8 @@ URL: https://AOMediaCodec.github.io/av1-avif
 Shortname: av1-avif
 Editor: Cyril Concolato, Netflix, cconcolato@netflix.com
```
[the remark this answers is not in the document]
aklemets: done

index.bs
```diff
@@ -5,8 +5,8 @@ URL: https://AOMediaCodec.github.io/av1-avif
...
 Former Editor: Paul Kerr, Netflix, pkerr@netflix.com
-Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of a [[!MIAF]] profile.
-Date: 2018-08-29
+Abstract: This document specifies syntax and semantics for the storage of [[!AV1]] images in the generic image file format [[!HEIF]], which is based on [[!ISOBMFF]]. While [[!HEIF]] defines general requirements, this document also specifies additional constraints to ensure higher interoperability between writers and readers when [[!HEIF]] is used with [[!AV1]] images. These constraints are defined in the form of profiles that extend the Multi-Image Application Format [[!MIAF]].
+Date: 2018-09-19
```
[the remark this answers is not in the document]
aklemets: done

index.bs
```diff
@@ -94,13 +103,17 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
-</ul>
+When an image item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and shall obey the following constraints:
+    
+  - The [=AV1 Image Item=] is a coded image item in which the coded image is a single valid [=AV1 Sample=] as defined in [[!AV1-ISOBMFF]].
+  - The [=AV1 Image Item=] shall be associated with an [=AV1 Item Configuration Property=].
 
-</p>
+The [=AV1 Sample=] that is the coded image of the [=AV1 Image Item=], is called <dfn>AV1 Image Item Data</dfn>. 
+The following constraints apply to [=AV1 Image Item Data=]:
+  - The [=AV1 Image Item Data=] shall be signaled as a sync sample, as defined in [[!AV1-ISOBMFF]].
```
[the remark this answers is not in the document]
aklemets: I did not mean to reference "signaling". I just meant to say that it has to be an AV1 sync sample. I have updated the text accordingly.

index.bs
```diff
@@ -131,27 +144,72 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+    - The [=AV1 Image Item=] should have its <code>[=still_picture=]</code> flag set to 1.
+    - The [=AV1 Image Item=] should have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
+    - The track handler for an [=AV1 Image Sequence=] shall be <code>pict</code>.
+
+    <h2 id="alpha-images">Alpha Image Items and Sequences</h2>
 <p>An <dfn>AV1 Alpha Image Item</dfn> (respectively an <dfn>AV1 Alpha Image Sequence</dfn>) is an [=AV1 Image Item=] (respectively [=AV1 Image Sequence=]</dfn>) with the following additional constraints:
-<ul>
-<li>The <code>[=mono_chrome=]</code> field in the [=Sequence Header OBU=] SHALL be set to 1.</li>
-</ul>
+    - The <code>[=mono_chrome=]</code> field in the [=Sequence Header OBU=] SHALL be set to 1.
```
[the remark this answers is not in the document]
aklemets: Fixed, plus in another place as well.

index.bs
```diff
@@ -131,27 +144,72 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
 
 <p>In [[!ISOBMFF]] or [[!HEIF]] files, if an [=AV1 Alpha Image Item=] (respectively. an [=AV1 Alpha Image Sequence=]) is used, the <code>aux_type</code> field of the <code>AuxiliaryTypeProperty</code> (respectively <code>AuxiliaryTypeInfoBox</code>) SHALL be set to <code>urn:mpeg:mpegB:cicp:systems:auxiliary:alpha</code>, as defined in [[!MIAF]].</p>
 
-<h2 id="AVIF-baseline-profile">MIAF AV1 Baseline Profile</h2>
+  <h2>Brands and file extensions</h2>
+
+<h3>AVIF image and image collection brand</h3>
+Files conformant with the brand-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) shall include the brand
```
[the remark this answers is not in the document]
aklemets: I realy meant to say "profile-independent", so I changed the text that way. Also added text saying that for "avis" the file is supposed to have one or more image sequences.

index.bs
```diff
@@ -115,13 +128,13 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
-<li>The values of the fields in the [=AV1CodecConfigurationBox=] SHALL match those of the [=Sequence Header OBU=] in the item data and in the [=AV1CodecConfigurationBox=], if any.</li>
-<li>[=Metadata OBUs=], if present, SHALL match the values given in other item properties, such as the PixelInformationProperty or ColourInformationBox.</li>
-</ul>
-This property SHALL be marked as essential.
-</p>
+
+  - [=Sequence Header OBUs=] should not be present in the [=AV1 Item Configuration Property=].
+  - The values of the fields in the [=AV1CodecConfigurationBox=] shall match those of the [=Sequence Header OBU=] in the [=AV1 Image Item Data=].
+  - [=Metadata OBUs=], if present, shall match the values given in other item properties, such as the PixelInformationProperty or ColourInformationBox.
+  - If a [=Sequence Header OBU=] is present in the [=AV1CodecConfigurationBox=], it shall match the  [=Sequence Header OBU=] in the [=AV1 Image Item Data=].
```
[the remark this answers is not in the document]
aklemets: Done. Also made sure that we refer to "AV1ConfigurationBox" when we mean that, rather than "AV1 Item Configuration Property" which was creating a kind of cyclical reference.

index.bs
```diff
@@ -131,27 +144,72 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
 
-<h2 id="AVIF-baseline-profile">MIAF AV1 Baseline Profile</h2>
+  <h2>Brands and file extensions</h2>
+
+<h3>AVIF image and image collection brand</h3>
+Files conformant with the brand-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) shall include the brand
+<dfn value="" export="" for="AVIF Image brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
+<p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
+<p>If 'avif' is specified in the major_brand field of the FileTypeBox, the file extension should be ".avif".
+The MIME Content Type for files with the ".avif" file extension shall be "image/avif".</p>
```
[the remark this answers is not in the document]
aklemets: Changed the two instances to "should". We could have a separate section dedictated to MIE types, but let's save that for a separate Pull Request.

index.bs
aklemets: Done

index.bs
aklemets: Done, but we can add the new section later.

aklemets:
@cconcolato -- I have pushed new changes. I think it is good to merge now, but I am waiting for you to look it over again and say that you are no longer requesting changes to the Pull Request.

index.bs
```diff
@@ -94,13 +104,17 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
-<li>The item data SHALL be the content of a single valid [=AV1 Sync Sample=] as defined in [[!AV1-ISOBMFF]] with the additional constraints that there SHALL be only one [=Sequence Header OBU=] and it SHALL have its <code>[=still_picture=]</code> flag set to 1 and SHOULD have its <code>[=reduced_still_picture_header=]</code> flag set to 1.
-<li>The item SHALL be associated with an [=AV1 Item Configuration Property=].</li>
-</ul>
+When an image item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and shall obey the following constraints:
+    
+  - The [=AV1 Image Item=] is a coded image item in which the coded image is a single valid [=AV1 Sample=] as defined in [[!AV1-ISOBMFF]].
+  - The [=AV1 Image Item=] shall be associated with an [=AV1 Item Configuration Property=].
 
-</p>
+The [=AV1 Sample=] that is the coded image of the [=AV1 Image Item=], is called <dfn>AV1 Image Item Data</dfn>.
```
cconcolato: The term "Sample" is confusing when used this way. Please replace with:
"The content of an [=AV1 Image Item=] is called the <dfn>AV1 Image Item Data</dfn>. "

index.bs
```diff
@@ -94,13 +104,17 @@ ISSUE: Should this specification also define a structural brand stricter than 'm
...
-</ul>
+When an image item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and shall obey the following constraints:
+    
+  - The [=AV1 Image Item=] is a coded image item in which the coded image is a single valid [=AV1 Sample=] as defined in [[!AV1-ISOBMFF]].
+  - The [=AV1 Image Item=] shall be associated with an [=AV1 Item Configuration Property=].
 
-</p>
+The [=AV1 Sample=] that is the coded image of the [=AV1 Image Item=], is called <dfn>AV1 Image Item Data</dfn>. 
+The following constraints apply to [=AV1 Image Item Data=]:
+  - The [=AV1 Image Item Data=] shall be an AV1 sync sample, as defined in [[!AV1-ISOBMFF]].
```
cconcolato: If you want to have a reference to [=AV1 Sample=], here it is fine. You could replace with:
"The [=AV1 Image Item Data=] shall be identical to the content of an [=AV1 Sample=] marked as 'sync', as defined in [[!AV1-ISOBMFF]].

index.bs
aklemets: Done

index.bs
aklemets: @cconcolato 
I changed the text as you requested, but this is what the AV1-ISOBMFF spec says:

> If an AV1 Sample is signaled as a sync sample (in the SyncSampleBox or by setting sample_is_non_sync_sample to 0), it SHALL be a Random Access Point as defined in [AV1].

Since in HEIEF we don't use the SyncSampleBox for image items, we don't "signal" samples as 'sync' or "mark" them as 'sync'. Maybe we should use the term "Random Access Point" instead?
E.g., "The [=AV1 Image Item Data=] shall be identical to the content of an [=AV1 Sample=] that is a Random Access Point, as defined in [[!AV1]]."

Anyway, I have changed the wording the way you requested, but we may want to revisit it later.

PR #22, opened 2018-10-16
fix bikeshed warnings 

cconcolato:
Used bikeshed version https://github.com/tabatkins/bikeshed/commit/018be3f805d3272ce699a232398d6e3902d1ea8d

Fix ids, links (difference between a property surrounded by `'` and definitions) ...

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -46,15 +46,20 @@ url: http://iso.org/#; spec: HEIF; type: property;
     text: pict
     text: pixi
 
-url: http://iso.org/#; spec: ISOBMFF; type: property;
+url: http://iso.org/#; spec: ISOBMFF; type: dfn;
     text: compatible_brands
     text: FileTypeBox
     text: major_brand
+
+url: http://iso.org/#; spec: ISOBMFF; type: property;
+    text: sync
 
 url: http://iso.org/#; spec: MIAF; type: property;
     text: clli
     text: mdcv
     text: miaf
+
+url: http://iso.org/#; spec: MIAF; type: dfn;
     text: primary image
 
 url: https://aomediacodec.github.io/av1-isobmff/#; spec: AV1-ISOBMFF; type: dfn;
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -145,36 +149,36 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+    - The track handler for an [=AV1 Image Sequence=] shall be <code>pict</code>.
+
+    <h2 id="alpha-images">Alpha Image Items and Sequences</h2>
 <p>An <dfn>AV1 Alpha Image Item</dfn> (respectively an <dfn>AV1 Alpha Image Sequence</dfn>) is an [=AV1 Image Item=] (respectively [=AV1 Image Sequence=]</dfn>) with the following additional constraints:
     - The <code>[=mono_chrome=]</code> field in the [=Sequence Header OBU=] shall be set to 1.
 
 <p>In [[!ISOBMFF]] or [[!HEIF]] files, if an [=AV1 Alpha Image Item=] (respectively. an [=AV1 Alpha Image Sequence=]) is used, the <code>aux_type</code> field of the <code>AuxiliaryTypeProperty</code> (respectively <code>AuxiliaryTypeInfoBox</code>) shall be set to <code>urn:mpeg:mpegB:cicp:systems:auxiliary:alpha</code>, as defined in [[!MIAF]].</p>
 
-  <h2>Brands and file extensions</h2>
+  <h2 id="brands-and-file-extensions">Brands and file extensions</h2>
```
aklemets: The reason I did not add a "id" attribute to the headings is that this is one of the benefits of using Bikeshed instead of authoring directly in HTML. Bikeshed will automatically generate the "id" attribute based on the text in heading, and it can be used for cross-references.

cconcolato: It was giving me a warning that they were not defined. If we want to have short, readable anchor names for URLs, it's probably better to define them, e.g. to avoid 'avif' being repeated.

PR #25, opened 2018-10-16
add formal MIME types registrations

cconcolato:


index.bs
```diff
@@ -159,21 +159,19 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
-  <h2 id="brands-and-file-extensions">Brands and file extensions</h2>
+  <h2 id="brands">Brands, Internet media types and file extensions</h2>
 
 <h3 id="image-and-image-collection-brand">AVIF image and image collection brand</h3>
 Files that conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) should include the brand
 <dfn value="" export="" for="AVIF Image brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
 <p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
-<p>If <code>avif</code> is specified in the major_brand field of the FileTypeBox, the file extension should be ".avif".
-The MIME Content Type for files with the ".avif" file extension shall be "image/avif".</p>
+<p>If <code>avif</code> is specified in the major_brand field of the FileTypeBox, the file extension and Internet Media Type should be as defined in [[#avif-mime-definition]].</p>
```
aklemets: The new section says that ".avif", ".heif", or ".hif" is are all allowed valid file extensions, without stating a preference. The nuance that ".avif" is the preferred file extension is lost. Can you fix it (same issue also with the sequence file extension)?

index.bs
```diff
@@ -215,3 +213,91 @@ The following additional constraints apply only to [=AV1 Image Sequences=]:
 - The AV1 profile level shall be either Main Profile or High Profile.
 - The AV1 level for Main Profile shall be 5.1 or lower.
 - The AV1 level for High Profile shall be 5.1 or lower.
+
+  <h2 id="mime-registration">AVIF Media Type Registrations</h2>
+
+  <h3 id="avif-mime-definition">AVIF Image Item Media Type Registration</h2>
+This section registers a new media type, <code>"image/avif"</code> in conformance with BCP 13. The information in this section is being submitted to the Internet Engineering Steering Group (IESG) for review, approval, and registration with the Internet Assigned Numbers Authority (IANA).
```
aklemets: IMO, this does not belong here. This is a template form that we use when we register MIME types with IANA. We will submit this form to IANA and they will publish it on their web site. Their published copy is the normative version, and we can reference it here, if necessary. I know the other specs included the IANA registration template in the spec itself, but "two wrongs don't make a right", so to speak.

My preference would be to have a section that defines the MIME type, but it does not need to strictly adhere to the IANA registration template -- especially since this template leads to the problem that I mentioned in my other comment, that multiple file extensions are listed as "allowed", without stating which one of the allowed file extensions is the recommended one.

Another issue is that the IANA template introduces various references to other specs that are not listed in the References section of this spec: RFC 6381, RFC 3986, BCP 13, RFC 4337, etc. We either have to add those to the References section of this spec, or simply avoid the problem by rewriting this section so that it is not an exact copy of the IANA template.

cconcolato: I agree that the text should not be here once IANA has published it. I would follow what other specs have done, i.e. keep the registration in our spec until IANA has published it. We should not worry about the references, and yes we should fix the nuance that was lost.

cconcolato:
@aklemets I think I have addressed both of your comments. Can you have another look?

PR #26, opened 2018-10-16
Various editorial changes on brands

cconcolato:
removing commented text
restructuring brand requirements as bullet points
clarifying use of HEIF brands
adding examples

It has conflicts with the changes on MIME types. Will fix them once #25 is merged.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -37,11 +37,15 @@ THE OTHER MEMBER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 </div>
 
 <pre class="anchors">
-url: http://iso.org/#; spec: HEIF; type: property;
+url: http://iso.org/#; spec: HEIF; type: dfn;
     text: coded image
     text: coded image item
-    text: colr
     text: image item
+
+url: http://iso.org/#; spec: HEIF; type: property;
+    text: colr
+    text: mif1
+    text: msf1
     text: pasp
     text: pict
     text: pixi
@@ -50,6 +54,7 @@ url: http://iso.org/#; spec: ISOBMFF; type: dfn;
     text: compatible_brands
     text: FileTypeBox
     text: major_brand
+    text: TrackTypeBox
 
 url: http://iso.org/#; spec: ISOBMFF; type: property;
     text: sync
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -162,17 +156,21 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
   <h2 id="brands-and-file-extensions">Brands and file extensions</h2>
 
 <h3 id="image-and-image-collection-brand">AVIF image and image collection brand</h3>
-Files that conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) should include the brand
-<dfn value="" export="" for="AVIF Image brand">avif</dfn> in the [=compatible_brands=] field of the [=FileTypeBox=].
-<p>Files should also carry a compatible brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.</p>
-<p>If <code>avif</code> is specified in the major_brand field of the FileTypeBox, the file extension should be ".avif".
+Files that conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]] and [[#alpha-images]]) should include in the [=compatible_brands=] field of the [=FileTypeBox=]:
+- the brand <dfn value="" export="" for="AVIF Image brand">avif</dfn> ,
+- the brand 'mif1' as recommended in [[!HEIF]],
```
aklemets: Doesn't "mif1" need to be enclosed in a "code" tag to avoid the Bikeshed dangling reference warning?

index.bs
```diff
@@ -190,7 +188,7 @@ The MIME Content Type for files with the ".avifs" file extension shall be "image
 
 This section defines the MIAF AV1 Baseline profile of [[!HEIF]], specifically for [[!AV1]] bitstreams, based on the constraints specified in [[!MIAF]] and identified by the brand  <dfn value export for="AV1 Image Item Type">MA1B</dfn>.
 
-If the brand <code>MA1B</code> is in the list of [=compatible_brands=] of a TrackTypeBox or [=FileTypeBox=], the common constraints in the Profiles and brands section shall apply.
+If the brand <code>MA1B</code> is in the list of [=compatible_brands=] of a [=TrackTypeBox=] or [=FileTypeBox=], the common constraints in the Profiles and brands section shall apply.
```
aklemets: This sentence mentions "Profiles and brands section". That could be changed into a hyperlink to that section.

index.bs
cconcolato: I already added 'mif1' in the HEIF-defined property so Bikeshed does not complain. The code tag is used only to change the style of the text and does not change Bikeshed processing IIUC.

There is some history here: Bikeshed was written for CSS and I think in the CSS specs everything that was surrounded by single-quote was a property. That's why when we use single-quotes we have to link using the "property" type.

PR #30, opened 2018-11-09
clarify support for multi-layer images (#18)

### LINKED ISSUE #18: Multi-layer images
Current spec is not clear about the status of scalable images. An item is said to be a AV1 Sync Sample but this is allowed to contain multiple frames from different layers. This should probably be made explicit and maybe restricted in profiles

cconcolato:
closes #18

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -104,7 +104,7 @@ This specification reuses syntax and semantics used in [[!AV1-ISOBMFF]].
 When an image item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn>AV1 Image Item</dfn>, and shall obey the following constraints:
   - The [=AV1 Image Item=] shall be associated with an [=AV1 Item Configuration Property=].
   - The content of an [=AV1 Image Item=] is called the <dfn>AV1 Image Item Data</dfn> and shall obey the following constraints:
-    - The [=AV1 Image Item Data=] shall be identical to the content of an [=AV1 Sample=] marked as 'sync', as defined in [[!AV1-ISOBMFF]].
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -136,6 +136,8 @@ In addition to the Image Properties defined in [[!HEIF]], such as 'colr', 'pixi'
 
 In general, it is recommended to use properties instead of [=Metadata OBUs=] in the [=AV1 Item Configuration Property=].
 
+For multi-layer images, the LayerSelectorProperty may be used and if used, it shall be as defined in [[!HEIF]].
```
aklemets: Should we say "may be _present_" instead of "may be _used_"? I think the term "used" is a bit vague -- it is not clear if it refers to the encoder or the parser/reader. By changing the word to "present" I think it becomes clear that the encoder may add this box. We don't say anything about whether a parser/reader must support the box or not, which I think is OK.

cconcolato: I agree my proposal is a bit vague. Actually, I wasn't clear about the use of the LayerSelectionProperty, that's probably why. Reading the HEIF spec again, I see:

> The decoding of a multi-layer image item may result into one or more reconstructed images.

> NOTE 1: For example, when a multi-layer image contains two views, a decoder is typically specified to decode and return the decoded images of both the views.

> When the decoding of a multi-layer image item results into more than one reconstructed images, a LayerSelectorProperty item property shall be present for the image item.

What I understand is:
- if you have a multi-layered image but in the traditional sense (i.e. the next layer improves the quality/resolution of the previous image), then you don't have to put the LayerSelectorProperty.
- if you have a multi-layered image for the multi-view case, you have to put the LayerSelectorProperty.

Is that correct @dwsinger?

The question for AV1 is for AV1: Is the notion of producing multiple reconstructed images applicable? 
I think it is, but I'm not sure. @agrange ? Can we have a base layer used by 2 enhancement layers that are not related to each other, but only predicting from the base?

agrange: Yes, there can be multiple unrelated enhancement layers that refer to the same base layer.

PR #31, opened 2018-11-09
introduce brand for intra-only sequences

### LINKED ISSUE #27: Consider defining a brand to identify intra-only sequences
As discussed last week, we should consider whether we want to define a brand to identify when image sequences are intra-only coded.

As of today, you have to inspect the track and detect that all samples are 'sync' samples. This might be a bit late and could result in files being downloaded while not being playable, if the decoder only supports intra. 

It might be better defined also within MPEG, as this is codec-independent.

cconcolato:
closes #27

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -161,13 +161,15 @@ Files that conform with the profile-independent restrictions in this document (s
 
 <p>If <code>avif</code> is specified in the <code>major_brand</code> field of the [=FileTypeBox=], the file extension and Internet Media Type should respectively be ".avif" and "image/avif" as defined in [[#avif-mime-definition]].</p>
 
-<h3 id="image-sequence-brand">AVIF image sequence brand</h3>
+<h3 id="image-sequence-brand">AVIF image sequence brands</h3>
 Files which contain one or more image sequences, and which conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]], [[#image-sequences]] and [[#alpha-images]]), should include in the [=compatible_brands=] field of the [=FileTypeBox=]:
 - the brand <dfn value="" export="" for="AVIF Image Sequence brand">avis</dfn>,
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -161,14 +161,16 @@ Files that conform with the profile-independent restrictions in this document (s
...
-<h3 id="image-sequence-brand">AVIF image sequence brand</h3>
+<h3 id="image-sequence-brand">AVIF image sequence brands</h3>
 Files which contain one or more image sequences, and which conform with the profile-independent restrictions in this document (sections [[#image-item-and-properties]], [[#image-sequences]] and [[#alpha-images]]), should include in the [=compatible_brands=] field of the [=FileTypeBox=]:
 - the brand <dfn value="" export="" for="AVIF Image Sequence brand">avis</dfn>,
 - the brand 'msf1' as recommended in [[!HEIF]],
 - a brand to identify the AVIF profile (see section [[#profiles]]), if any, with which the file complies.
 
 <p>If <code>avis</code> is specified in the major_brand field of the FileTypeBox, the file extension and Internet Media Type should be respectively ".avifs" and "image/avif-sequence" as defined in [[#avis-mime-definition]].</p>
 
+Additionally, if the image sequences are made only of [=AV1 Samples=] marked as 'sync', then the brand <dfn value="" export="" for="AVIF Intra-only brand">avio</dfn> should be used.
```
aklemets: This works, but it would be nicer to have a separate section (5.3) for the intra only brands. The existing section (5.2) could be renamed "AVIF generic image sequence brand" and the new section would be titled "AVIF intra-only image sequence brand".
Also, Examples 1 and 2 in the profiles section could be augmented to give an example of how it looks like when the new 'avio' brand is used.

cconcolato: I have a slight preference for having only one section, as `avio` brand is really linked to `avis`.
I modified the example to add `avio`.

PR #32, opened 2018-12-16
Add some AVIF test files

aklemets:
Uploading some .avif test files. I don't have a public server where to put these files, so just putting them in a sub folder in the same repro as the AVIF spec itself.


cconcolato:
Thanks @aklemets. I noticed 2 small issues:
- In Chimera_8bit_cropped_480x256.avif (and the 10 bits version too), the 'colr' box is using 0 for `transfer_characteristics`. It's a reserved value. Shouldn't it be 2? Or maybe the box could be removed?
- In Chimera_10bit_cropped_to_1920x1008.avif (and the one with metadata), the pixi box indicates that the channels are 8 bits.

FYI, I had to update mp4box.js: some boxes irot, mdcv, clli were not implemented. Also the JS player does not work with the 444 sequence. BTW it would not support a 4:2:2 either and probably not monochrome. We should try.

Also do you know if you sequences use the reduce still header?

aklemets:
I pushed new changes where I have fixed the 'pixi' and 'colr' boxes for the affected files. I made sure to use "2" in the 'colr' box instead of "0". I cannot remove the 'colr' box because if this box is absent it implies certain defaults: BT.709 video primaries, and so on.

The cause of the problem is that the original video stream (the Chimera files) have color_description_present_flag set to 0 in the Sequence Header OBU. So, that means that color space info is not specified in the Sequence Header, and the AV1 spec does not specify a default color space as far as I can tell. So, the color information is truly unknown, so I am forced to add a 'colr' box to the HEIF file.

In contrast, the AV1 streams that I downloaded from Elecard do include color information, and the values match the defaults (BT.709. etc.) so I am able to omit the 'colr' box, which I do whenever it is possible.

PR #50, opened 2019-01-31
adding sequence with alpha and video

### LINKED ISSUE #35: Create conformance images and image sequences with alpha channel

cconcolato:
closes #35 

The file contains a pict track, an auxv track linked to it, a primary item and an alpha item linked to it.


cconcolato:
Thanks for the detailed review! I will update/fix the files.

cconcolato:
@aklemets I updated the file:
1. `pixi` is added to the image items. It was missing from MP4Box. I added it, see https://github.com/gpac/gpac/pull/1201. I will update the other files as well.
2. `auxC` is added. There was a bug in MP4Box
3. `auxi` is added to the alpha video sequence. I also added an option for that to MP4Box.
4. Indeed, the HEIF spec says:
> the track_in_movie flag in TrackHeaderBox of auxiliary image sequences tracks should be equal to 0

It's only a should. I did not have the time to change it. The flags are currently set to 1 in both tracks because the ISOBMFF spec says:
> If in a presentation all tracks have neither track_in_movie nor
track_in_preview set, then all tracks shall be treated as if both flags were set on all tracks.

I'll see if I can update that in the future.

5. I changed the ID to avoid the clash but I think the clash would only be relevant if there was a `GroupsListBox` in the file, as HEIF says:
>When GroupsListBox is present in a file-level MetaBox, there shall be no item_ID value in ItemInfoBox in any file-level MetaBox that is equal to the track_ID value in any TrackHeaderBox.

6. I had put the track reference but indeed it was in the wrong direction (from the main to the alpha)
7. changed.

Let me know if this works for you.

aklemets:
Something happened with the Chimera file. Now the still image item has id no. 2, which is the same ID as the track. Previously, it did not have this problem. However, in the alpha_video file, the IDs are all unique now. The rest looks fine.

cconcolato:
Restored the id.

PR #65, opened 2019-10-18
Update Netflix testFiles

joedrago:
* Correct CICP values in BT2020 derivatives into nclx box (to use NCL instead of CL)
* Put correct CICP values into OBU payload in addition to nclx
* Add a few additional frames


cconcolato:
LGTM. I compared the old and new files and confirm that from a box/file format point of view, only the nclx values have changed in HDR images.

PR #99, opened 2020-09-03
Add constraint that all frames in a sequence have the same width and height

### LINKED ISSUE #75: How about additional constraints: "All frames in Image Sequence file should have the same size (widthxheight)"
Hi. I am currently supporting AVIF/AVIFS images in Chrome, and implementing AVIFS in [cavif](https://github.com/link-u/cavif) and [davif](https://github.com/link-u/davif).

According to Wan-Teh, a chrome developer, it is required to all frames in an animated image to be the same size in Chrome(widthxheight).

I think it's reasonable constraints.

FYI: In WebP, it looks all frame have the same size.
https://github.com/webmproject/libwebp/blob/master/src/webp/mux.h#L277-L312

cconcolato:
Closes #75

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/99.html" title="Last updated on Sep 3, 2020, 2:59 PM UTC (04816af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/99/66e9d59...04816af.html" title="Last updated on Sep 3, 2020, 2:59 PM UTC (04816af)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -124,8 +124,9 @@ url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn
     text: Sequence Header OBU
     text: Metadata OBU
     text: Temporal Unit
-
-url: http://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn;
+    text: Frame
+    text: FrameWidth
+    text: FrameHeight
     text: mono_chrome
     text: still_picture
     text: reduced_still_picture_header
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -124,8 +124,9 @@ url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn
     text: Sequence Header OBU
     text: Metadata OBU
     text: Temporal Unit
-
-url: http://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1; type: dfn;
```
dwbuiten, line 128: Stray change?

cconcolato: No. This line is (almost) a duplicate of line 122, except that it uses http instead of https. I think initially we had 2 sections to distinguish AV1-defined words (e.g. Temporal Unit) from AV1-defined keywords (e.g. mono_chrome). I'm not sure we need to make this distinction. WDYT?

leo-barnes:
The issue this is based on was closed without action since ISOBMFF specified how it should be handled. Do we still want to add the restriction? Can't remember what we decided in the meetings.

cconcolato:
Closing this PR as it was agreed to close the issue without action.

PR #112, opened 2020-10-22
Remove wrong recommendations regarding still flags in image sequences

### LINKED ISSUE #107: AV1 Image sequence section needs fix regarding still_picture flag and reduce_still_picture header flag
The current text in https://aomediacodec.github.io/av1-avif/#image-sequences says:
> * The AV1 Image Item should have its still_picture flag set to 1.
> * The AV1 Image Item should have its reduced_still_picture_header flag set to 1.

First, the section is about image sequences so the notion of AV1 Image Item does not make sense. 
Second, given that these flags are at the Sequence Header OBU, it does not make sense either. 
Third, it is not clear what the intent is because recommending still_picture to 1 means that the sequence has only 1 frame so to have a real image sequence you'd have to have multiple AV1 Sequences, one per sample.

It is proposed to remove these 2 bullet points.

cconcolato:
closes #107

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -193,8 +193,6 @@ For multi-layered images, the LayerSelectorProperty may be present as defined in
 <p>
     An <dfn>AV1 Image Sequence</dfn> is defined as a set of AV1 [=Temporal Units=] stored in an [=AV1 track=] as defined in [[!AV1-ISOBMFF]] with the following constraints:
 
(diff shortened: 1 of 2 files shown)


leo-barnes:
Looks good to me.

leo-barnes (approved):

wantehchang (approved):
LGTM. R=wtc. Thank you for making the change, Cyril!

Issue #104, opened 2020-09-21
AVIF should give examples and recommendations on how multi-layer/progressive coding should be done


leo-barnes:
The conclusion from the latest meetings and discussions on the list is that there are multiple ways of doing progressive coding in AVIF. The best ways involve multi-layer coding, but it's not completely obvious exactly how it should be done. The following additions to AVIF would be good:
1. Some pointers on when 'lsel' should be used to expose layers at the container level. Always? Only for larger images where you may want to decode at lower resolution?
2. Recommended structure of a file for optimal interoperability and progressive decoding. Do we want to recommend a specific layering structure?
3. Examples
4. A decoder model that has recommended behavior for how a decoder should handle progressive decoding of the given examples.

PR #118, opened 2020-10-29
Improve use of layered images.

### LINKED ISSUE #103: Semantics of layer_id in LayerSelectorProperty needs to be specified
HEIF indicates:
> layer_id specifies the layer identifier of the image among the reconstructed images that is
described by subsequent descriptive item properties in the item property association order
and manipulated by transformative item properties, if any, to generate an output image of
the image item. The semantics of layer_id are specific to the coding format and are
therefore defined for each coding format for which the decoding of a multi-layer image item
can result into more than one reconstructed images.

I would propose to clarify that `layer_id` in AVIF means:
- either the 0-based index of the desired operating point in the list of operating points 
- or the actual operating_point_idc to be used

I don't think it make sense to point to a specific layer as they could be non decodable on their own.

This is related to #18 and #40

cconcolato:
closes #103

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/118.html" title="Last updated on Feb 3, 2021, 7:49 PM UTC (e6727ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/118/13bbd2c...e6727ba.html" title="Last updated on Feb 3, 2021, 7:49 PM UTC (e6727ba)">Diff</a>


leo-barnes (changes requested):
Looks good. Found a typo which I commented on inline. Do we also want to add recommendations, or do we do that as part of #104 ?

leo-barnes (approved):
Looks good to me. My only concern is for the case where you have two output images (for instance two viewpoints) sharing a base layer. If you don't use `lsel`, a decoder may try to use the layers for progressive refinement which is going to look a bit weird. On the other hand, the spec does note that if you want full control over what should happen you should be using `lsel`.

index.bs
```diff
@@ -170,19 +182,68 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
...
+
+In an [=AV1 Multi-Layered Image Item=], a spatial layer may improve the resolution or quality of the image decoded based on one or more of the previous layers. A layer may also provide an image that does not depend on the previous layers. Not all layers are expected to produce an image meant to be rendered. Some decoded images may be used only as intermediate encodes. The [=Sequence Header OBU=] defines the combinations of layers that are expected to produce output images. Each combination is called an [=Operating Point=]. An [=AV1 Image Item=] may contain data from multiple [=Operating Points=]. It may have [=Operating Points=] representing the same image in multiple resolutions or qualities. It may have [=Operating Points=] representing different images encoded based on a common set of layers.
```
wantehchang, line 202: Cyril: In the sentence

> Some decoded images may be used only as intermediate encodes.

did you mean "intermediate encodes" or "intermediate decodes"?

index.bs
```diff
@@ -170,19 +182,68 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
...
+The 'lsel' property defined in [[!HEIF]] may be associated to an [=AV1 Multi-Layered Image Item=].
+
+If an [=AV1 Multi-Layered Image Item=] is associated with an 'lsel' property, the 'layer_id' value shall be between 0 and [=operating_points_cnt_minus_1=]. The 'ispe' property shall describe the dimension of the output image of the corresponding [=Operating Point=]. Renderers should use the 'layer_id' value as the return of the [=choose_operating_point=] function and should not render any other output image corresponding to other [=Operating Points=]. If the [=Operating Point=] corresponding to the 'layer_id' value exceeds the renderer's decoding capabilities, the renderer should not process the item.
+
+If an [=AV1 Multi-Layered Image Item=] is not associated with an 'lsel' property, the 'ispe' property shall describe the dimension of the output image of the operating point whose index is 0. Renderers are free to render either: only the output image of the [=Operating Point=] matching best its decoding capabilities, or to render all output images of all the [=Operating Points=] processed, resulting in a form of progressive decoding.
```
wantehchang, line 210: Nit: I think "best matching" sounds better than "matching best".

index.bs
```diff
@@ -170,19 +182,68 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
...
+
+<h4 id ="layered-properties">Properties</h4>
+<h5 id ="layer-selector-property">Use of Layer Selector Property</h4>
+The 'lsel' property defined in [[!HEIF]] may be associated to an [=AV1 Multi-Layered Image Item=].
+
+If an [=AV1 Multi-Layered Image Item=] is associated with an 'lsel' property, the 'layer_id' value shall be between 0 and [=operating_points_cnt_minus_1=]. The 'ispe' property shall describe the dimension of the output image of the corresponding [=Operating Point=]. Renderers should use the 'layer_id' value as the return of the [=choose_operating_point=] function and should not render any other output image corresponding to other [=Operating Points=]. If the [=Operating Point=] corresponding to the 'layer_id' value exceeds the renderer's decoding capabilities, the renderer should not process the item.
```
wantehchang, line 208: IMPORTANT: This paragraph and the next paragraph seem to assume that each operating point has only one output image. If I understand the AV1 spec correctly, an operating point may have an output image per layer. You can search for "shown frame" in the AV1 spec to find the description. Here is the relevant text in Section 7.5. Ordering of OBUs, pp. 204-205:

> If scalability is being used (OperatingPointIdc not equal to 0), then only a subset of frames are part of the operating point.
> For each operating point, the following constraints must hold:
> 
> - The first frame header that will be decoded must have frame_type equal to KEY_FRAME and show_frame equal to 1.
> 
> - Every layer that has a coded frame in a temporal unit must have exactly one shown frame that is the last frame of that layer in the temporal unit.

The AV1 spec describes an obscure, unnamed application function that determines whether an output frame will actually be ouput, in Section 7.18 (Output process), page 336:

> If scalability is being used (OperatingPointIdc not equal to 0), an application-specific function is called to decide whether this frame will be output. If this function returns a value equal to 0, then this process terminates immediately.
> **Note:** Applications that are displaying the decoded video are expected to only display one frame from each
temporal unit within the selected operating point. This frame should be the highest spatial layer that is both within the operating point and present within the temporal unit. Other applications may set their own policy about which frames are output.

I think this unnamed application-specific function can be used to control whether a multi-layered image will be rendered progressively or not. In all the AV1 decoders (dav1d, libaom, libgav1), this applicaton-specific function is implemented by the "output all layers" decoder setting.

index.bs
```diff
@@ -170,19 +182,68 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
...
+  Container:      <code>ItemPropertyContainerBox</code>
+  Mandatory:      No
+  Quantity:       One
+</pre>
+
+<h6 id ="layered-image-selector-property-description" class="no-toc">Description</h6>
+
+The <dfn>AV1LayeredImageIndexingProperty</dfn> provides the size in bytes of each layer in the image data. It enables determining the byte ranges required to process an [=Operating Point=].
+
+The [=AV1LayeredImageIndexingProperty=] property may be associated to an [=AV1 Multi-Layered Image Item=]. It shall not be marked as essential. It shall not be associated with [=AV1 Image Items=] that are not using multiple layers.
```
wantehchang, line 232: Nit: associated to => associated with

index.bs
```diff
@@ -170,19 +182,68 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
...
+
+NOTE: The dimensions of the final or intermediate output images might not match the ones given in the 'ispe' property. Renderers are expected to scale the output image to match the 'ispe' property.
+
+NOTE: When such a progressive decoding is not desired (e.g. for multi-view images) or when a controlled rendering of the different output images is desired (e.g. grid of multi-layered images), multiple [=AV1 Image Items=] sharing the same [=AV1 Image Item Data=] can be created and associated with different 'lsel' properties, each with a different value of 'layer_id'. In the case of an image grid composed of 4 layered images, each made of 2 operating points, controlled rendering could be achieved by creating a set of items as follows: 4 items each corresponding to the low quality/resolution image of the first operating point of the 4 multi-layered images, i.e. with 'layer_id' set to 0; 4 items corresponding the second operating point, i.e. with 'layer_id' set to 1; a grid item composing the first set of 4 images; another grid item composing the second set of 4 images.
```
wantehchang, line 214: (Note: This paragraph also assumes that each operating point has only one output image, which I believe differs from the AV1 spec.)

The AV1 spec says:

> **choose_operating_point( )** is a function call that indicates that the operating point should be selected.
> 
> The implementation of this function depends on the capabilities of the chosen implementation. The order of operating points indicates the preferred order for producing an output: a decoder should select the earliest operating point in the list that meets its decoding capabilities as expressed by the level associated with each operating point.

Therefore, an earlier operating point should have a higher quality/resolution than a later operating point. This example says;

> 4 items each corresponding to the low quality/resolution image of the first operating point of the 4 multi-layered images, i.e. with 'layer_id' set to 0;

I think "low" should be changed to "high".

index.bs
```diff
@@ -170,19 +182,68 @@ When an image item is of type <dfn value export for="AV1 Image Item Type">av01</
...
+The <dfn>AV1LayeredImageIndexingProperty</dfn> provides the size in bytes of each layer in the image data. It enables determining the byte ranges required to process an [=Operating Point=].
+
+The [=AV1LayeredImageIndexingProperty=] property may be associated to an [=AV1 Multi-Layered Image Item=]. It shall not be marked as essential. It shall not be associated with [=AV1 Image Items=] that are not using multiple layers.
+
+<h6 id ="layered-image-selector-property-syntax" class="no-toc">Syntax</h6>
+
+```
+class AV1LayeredImageIndexingProperty extends ItemFullProperty('a1lx', version = 0, flags) {
+    FieldLength = ((flags & 1) + 1) * 16;
+    unsigned int(FieldLength)[4] layer_size;
```
wantehchang, line 239: I guess the offset is implicitly the sum of all previous layer sizes?

index.bs
cconcolato: The sentence is mixing both and is not correct, but depending on the point of view, the intermediate images can be viewed as intermediate encodes (from an encoder's point of view) or intermediate decodes (from a decoder's point of view).

index.bs
cconcolato: Thank you. I need to study more then.

index.bs
```diff
@@ -188,15 +194,56 @@ In addition to the Image Properties defined in [[!HEIF]], such as 'colr', 'pixi'
...
+
+If an [=AV1 Multi-Layered Image Item=] is not associated with an 'lsel' property, the 'ispe' property shall describe the dimension of the output image of the operating point whose index is 0. Renderers are free to render either: only the output image of the [=Operating Point=] matching best its decoding capabilities, or to render all output images of all the [=Operating Points=] processed, resulting in a form of progressive decoding.
+
+NOTE: The dimensions of the final or intermediate output images might not match the ones given in the 'ispe' property. Renderers are expected to scale the output image to match the 'ispe' property.
```
leo-barnes, line 212: Why would the dimensions of the final output image not match the 'ispe' property? I can understand why the intermediates might not, but I would have expected the 'ispe' to match the final output.

index.bs
leo-barnes: This is explicitly tailored for AV1. Do we want to make it slightly more generic so that it can also work for AV2?

Do we want the array to be variable size? Doing so would give two things:
1. You would only have to specify the layers that actually exist.
2. It would be future proof with AV2 if AV2 adds more layers than 4.

leo-barnes (approved):
Looks good to me, minor comments inline.
Sorry for the delay, must have missed the notification that you updated the review.

index.bs
cconcolato: It's an option. We can always define a new version of the box for AV2 as we de don't know how AV2 will be.

wantehchang: Adding an array size field will make it clear which array entries are valid, without referring to the OperatingPointIdc bitmask. I guess it will also save a few bytes in the common cases.

wantehchang: Cyril: I just realized that AV1 does not seem to require that a bitstream use contiguous spatial_id values. So it may be valid for a bitstream with two spatial layers to use spatial_id values 0 and 2. So I think your current design is good. Unused spatial_id values can be indicated by 0 in the layer_size array.

leo-barnes: > Cyril: I just realized that AV1 does not seem to require that a bitstream use contiguous spatial_id values. So it may be valid for a bitstream with two spatial layers to use spatial_id values 0 and 2. So I think your current design is good. Unused spatial_id values can be indicated by 0 in the layer_size array.

👍 Sounds like a strong justification for the current design.

index.bs
leo-barnes: Just listened through the recording of the meeting.

If multiple outputs per operating point is possible like suggested by Wan-Teh and we let 'lsel' select a specific layer rather than operating point:
1. We need to clarify which operating point should be used if there are multiple. Or we can make it simple and say that operating point 0 shall always be used.
2. We need to clarify which operating point (if we don't always force 0) and layer the 'ispe' describes if the 'lsel' is not present

If multiple outputs per operating point is not possible, the `NOTE` above is slightly confusing. In the paragraph above it, we mention that if there is no 'lsel' the 'ispe' shall match the output of operating point 0. I think that what the note is saying is that if the decoder selects another operating point, the dimensions of the final output may not match the 'ispe' and that it should scale the image to that size if needed.

index.bs
leo-barnes, line 232: Do we want to make the property more than just descriptive? We could make the presence of it indicate that the encoder thinks progressive refinement should be done if applicable.

Something like this:
`[lsel,a1lx]`: Decoder should decode the operating point/layer specified by `lsel` and do progressive refinement (if applicable) if that has not yet been downloaded
`[a1lx]`: Same as `[lsel,a1lx]` with an implicit `lsel` = 0
`[lsel]`: Decoder should decode the operating point/layer specified by `lsel`. Progressive refinement should **not** be done. (The other layers may not be the same image.)
`[]`: Same as `[lsel]` with an implicit `lsel` = 0

index.bs
cconcolato: Fixed

index.bs
cconcolato: Fixed

index.bs
cconcolato: Fixed.

index.bs
cconcolato: I proposed a rewrite of the section which I think adresses the concern.

index.bs
cconcolato: The latest rewrite tries to address this comment.

index.bs
leo-barnes: Conclusion of the meeting was that this would probably just complicate things.

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+
+[[!AV1]] delegates the selection of which [=Operating Point=] to process to the application. AVIF defines the [=OperatingPointSelectorProperty=] to control this selection. In the absence of an [=OperatingPointSelectorProperty=] associated with an [=AV1 Image Item=], the AVIF renderer is free to process any [=Operating Point=] present in the [=AV1 Image Item Data=]. In particular, when the [=AV1 Image Item=] is composed of a unique [=Operating Point=], the [=OperatingPointSelectorProperty=] should not be present. If an [=OperatingPointSelectorProperty=] is associated with an [=AV1 Image Item=], the <code>op_index</code> field indicates which [=Operating Point=] is expected to be presented for this item.
```
leo-barnes, line 206: `field indicates which [=Operating Point=] is expected to be presented for this item.`
Do we want to make this a shall? If the file creator selected a specific operating point due to the other operating points containing completely different images, not decoding the specified operating point could get very weird.

Same goes the `lsel` wording.

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+    u8 op_index;
+}
+```
+
+<h6 id ="operating-point-selector-property-semantics" class="no-toc">Semantics</h6>
+
+<dfn noexport>op_index</dfn> indicates index of the operating point to be processed for this item. Its value shall be between 0 and [=operating_points_cnt_minus_1=].
+
+<h5 id ="layer-selector-property">Layer Selector Property</h5>
+The 'lsel' property defined in [[!HEIF]] may be associated with an [=AV1 Image Item=]. The 'layer_id' indicates the value of the spatial id to render. The corresponding spatial layer shall produce an output frame. Other layers may be needed to decode the indicated layer.
```
leo-barnes, line 248: Do we need to explicitly state that `layer_id` should be within a certain range or something like that (like how we do for `a1op` and `operating_points_cnt_minus_1`)?

It's probably not needed, but thought I'd check just to be sure.

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+  Box Type:       <dfn value export for="AV1LayeredImageIndexingProperty">a1lx</dfn>
+  Property type:  Descriptive item property
+  Container:      <code>ItemPropertyContainerBox</code>
+  Mandatory:      No
+  Quantity:       One
+</pre>
+
+<h6 id ="layered-indexing-property-description" class="no-toc">Description</h6>
+
+The <dfn>AV1LayeredImageIndexingProperty</dfn> property may be associated with an [=AV1 Image Item=]. It documents the size in bytes of each layer in the [=AV1 Image Item Data=], and enables determining the byte ranges required to process an [=Operating Point=]. If associated, it shall not be marked as essential.
```
leo-barnes, line 268: `and enables determining the byte ranges required to process an [=Operating Point=]`
Should this be changed to `required to process a specific layer` now that we have both `a1op` and `lsel`?

leo-barnes (approved):
I think this looks good. Some minor questions inline.

index.bs
cconcolato: I refrain from 'shall' statements on player/reader behaviors and keep them for files (i.e. can be validated by a tool), but this can be changed.

index.bs
cconcolato: It would be a good clarification, yes.

index.bs
cconcolato: That's a good question. Yes, you can determine the layer byte range but in combination with the Sequence Header you can also determine the operating point. Should we say `required to process a specific layer or an operating point`?

leo-barnes: Sounds good to me!

index.bs
leo-barnes: > I refrain from 'shall' statements on player/reader behaviors and keep them for files
You're right, that's probably the correct approach. It should be enough to explain what the expected player behaviour is. A HEIF box GUI viewer may want to show everything in the file, and we don't want to say that that is not allowed.

index.bs
cconcolato: done.

index.bs
cconcolato: done.

leo-barnes (approved):
👍

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+<h6 id ="operating-point-selector-property-semantics" class="no-toc">Semantics</h6>
+
+<dfn noexport>op_index</dfn> indicates index of the operating point to be processed for this item. Its value shall be between 0 and [=operating_points_cnt_minus_1=].
+
+<h5 id ="layer-selector-property">Layer Selector Property</h5>
+The 'lsel' property defined in [[!HEIF]] may be associated with an [=AV1 Image Item=]. The 'layer_id' indicates the value of the [=spatial_id=] to render. The value shall be between 0 and 3. The corresponding spatial layer shall be present in the bitstream and shall produce an output frame. Other layers may be needed to decode the indicated layer.
+
+If this property is associated with an item, the 'ispe' property documents the dimensions of the output frame produced by decoding this layer.
+
+NOTE: The dimensions of possible intermediate output images might not match the ones given in the 'ispe' property. Renderers are expected to scale the output image to match the 'ispe' property.
```
wantehchang, line 252: This NOTE seems to be in the wrong place. If the 'lsel' property is present, the layers not selected by the 'ispe' property will not be rendered. The renderer should not need to scale the output images of the unselected layers.

It seems that this NOTE should be moved to the previous subsection, at line 234, after the paragraph "In the absence of a 'lsel' property associated with the item".

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+
+NOTE: When such a progressive decoding of the layers within an [=Operating Point=] is not desired or when a controlled rendering of the different output images is desired (e.g. grid of layered images), multiple [=AV1 Image Items=] sharing the same [=AV1 Image Item Data=] can be created and associated with different 'lsel' properties, each with a different value of 'layer_id'. In the case of an image grid composed of 4 layered images, each made of 2 layers, controlled rendering could be achieved by creating a set of items as follows: 4 items each corresponding to the low quality/resolution image as decoded from the first layer of each of the 4 layered images, i.e. with 'layer_id' set to 0; 4 items corresponding the second layer, i.e. with 'layer_id' set to 1; a grid item composing the first set of 4 images; another grid item composing the second set of 4 images.
```
wantehchang, line 212: [Just a comment. No change requested.] I have to admit I don't understand the layered grid image example.

index.bs
wantehchang, line 206: [Just a comment. No change requested.] Just FYI: The `op_index` field here corresponds to the `operatingPoint` variable in the AV1 spec. There is also an `OperatingPointIdc` variable in the AV1 spec, but `OperatingPointIdc` is a bit mask, not an index. Assuming "Idc" stands for "index", I find the names used in the AV1 spec very confusing.

wantehchang (approved):
LGTM. I suggest a few small changes.

index.bs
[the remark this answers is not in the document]
cconcolato: No strong opinion here. I changed "presented" to "processed".

index.bs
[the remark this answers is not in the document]
cconcolato: A grid is composing the images or is composed of the images. Both are equivalent to me. I rephrased as suggested.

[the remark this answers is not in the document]
cconcolato: Let me try to explain so that we can come up with a better text. 

Let's say you have a grid with 4 images, each made of only one OP and each OP made of 2 layers (increased resolution or improved quality). If the author does not provide any additional information (no lsel), according to what is in the spec, the renderer is free to display each image in the grid the way it wants. This means it could display intermediate rendering of the grid where 2 images would show the low quality and 2 would show the high quality. If the author does not want that, it could: create 2 grids. The first grid would point to 4 items that have lsel = 1 and the second grid point to 4 different items that have lsel = 2. The 2 grids could be grouped as recommended by HEIF with 'altr' grouping. 

As I write this explanation, I realize this is too complicated for an example. I simplified the example (not mentioning grids).

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+  Quantity:       One
+</pre>
+
+<h6 id="operating-point-selector-property-description" class="no-toc">Description</h6>
+
+An <dfn>OperatingPointSelectorProperty</dfn> may be associated with an [=AV1 Image Item=] to provide the index of the operating point to be processed for this item. If associated, it shall be marked as essential.
+
+In the absence of a 'lsel' property associated with the item:
+- If an [=OperatingPointSelectorProperty=] is associated with an [=AV1 Image Item=], the 'ispe' property associated with this item shall describe the dimension of the output image corresponding to the operating point.
+- If it is not associated with an item, the 'ispe' documents the dimensions of the the operating point whose index is 0.
```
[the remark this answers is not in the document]
cconcolato: corrected as suggested

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+  Box Type:       <dfn value export for="AV1LayeredImageIndexingProperty">a1lx</dfn>
+  Property type:  Descriptive item property
+  Container:      <code>ItemPropertyContainerBox</code>
+  Mandatory:      No
+  Quantity:       One
+</pre>
+
+<h6 id ="layered-image-indexing-property-description" class="no-toc">Description</h6>
+
+The <dfn export>AV1LayeredImageIndexingProperty</dfn> property may be associated with an [=AV1 Image Item=]. It documents the size in bytes of each layer in the [=AV1 Image Item Data=], and enables determining the byte ranges required to process a specific layer or an [=Operating Point=]. If associated, it shall not be marked as essential.
```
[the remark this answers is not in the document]
cconcolato: You can use the index to get either one layer or all the layers a given operating point. I changed it to "to process one or more layers of an [=Operating Point=]"

index.bs
cconcolato: Even if you select a layer, you may have to decode intermediate images whose dimensions might differ but you're right that in this case the expectation is that you don't render these intermediate images so scaling should not be needed. I moved the note as suggested.

index.bs
cconcolato: See previous response. I simplified the example.

index.bs
cconcolato: Should the semantics of `op_index` indicate that it gives the value of the `operatingPoint` variable?

wantehchang: > Should the semantics of `op_index` indicate that it gives the value of the `operatingPoint` variable?

If it is not too verbose, we can mention that OperatingPointSelectorProperty implements the `choose_operating_point()` function and `op_index` gives the value of the `operatingPoint` variable in the AV1 spec. In other words, this is equivalent to the following statement in Section. 5.5.1 of the AV1 spec:

`    operatingPoint = choose_operating_point( )`

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+
+NOTE: When an author wants to offer the ability to render multiple [=Operating Points=] from the same AV1 image (e.g. in the case of multi-view images), multiple [=AV1 Image Items=] can be created that share the same [=AV1 Image Item Data=] but have different [=OperatingPointSelectorProperty=]s.
+
+[[!AV1]] expects the renderer to display only one frame within the selected [=Operating Point=], which should be the highest spatial layer that is both within the [=Operating Point=] and present within the temporal unit, but [[!AV1]] leaves the option for other applications to set their own policy about which frames are output. AVIF sets a different policy, and defines how the 'lsel' property (defined in [[!HEIF]]) is used to control which layer is rendered. In the absence of a 'lsel' property associated with an [=AV1 Image Item=], the renderer is free to render either: only the output image of the highest spatial layer, or to render all output images of all the intermediate layers, resulting in a form of progressive decoding. If a 'lsel' property is associated with an [=AV1 Image Item=], the renderer is expected to render only the output image for that layer.
```
wantehchang, line 210: The first sentence says "..., but [[!AV1]] leaves the option for other applications to set their own policy about which frames are output." We could note that this refers to the second paragraph in Section 7.18.1 of the AV1 spec:

```
If scalability is being used (OperatingPointIdc not equal to 0), an
application-specific function is called to decide whether this frame
will be output. If this function returns a value equal to 0, then this
process terminates immediately.
```

I think this cross reference would be useful because that "application-specific function" does not have a name and it is hard to find that paragraph when one reads the AV1 spec.

wantehchang, line 210: The sentence "In the absence of a 'lsel' property associated with an [=AV1 Image Item=], the renderer is free to render ..." implies that an AV1 Image Item with multiple layers is allowed to have no 'lsel' property associated with it. This seems to contradict the following requirement in HEIF Section 6.5.11.1:

```
If the decoding of a multi-layer image item results into more than one
reconstructed image, the 'lsel' item property shall be associated with
the image item. ...
```

Note: The paragraph "In the absence of a 'lsel' property associated with the item:" at line 231 also implies that an AV1 image item with multiple layers may have no 'lsel' property associated with it.

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+<h5 id ="operating-point-selector-property">Operating Point Selector Property</h5>
+
+<h6 id ="operating-point-selector-property-definition" class="no-toc">Definition</h6>
+
+<pre class="def">
+  Box Type:       <dfn value export for="OperatingPointSelectorProperty">a1op</dfn>
+  Property type:  Descriptive item property
+  Container:      <code>ItemPropertyContainerBox</code>
+  Mandatory:      No
+  Quantity:       One
```
wantehchang, line 224: Nit: In HEIF, the "Quantity" of this kind of item property is specified as "Zero or one". That makes it clearer that the item property can be absent.

wantehchang (approved):
Hi Cyril: The new version of the PR looks good! I noticed a contradiction with HEIF's requirement on whether the 'lsel' property may be absent for a layered image item.

index.bs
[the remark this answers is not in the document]
wantehchang: Thanks for the explanation! What I don't understand is whether we want to allow a layered grid image to be decoded (or rendered) progressively. This PR describes progressive decoding only when the 'lsel' property is absent. In this grid image example, the two grid images each have an associated 'lsel' property, so they are not expected to be decoded/rendered progressively. If that is intended, then it would be good to describe another example of a grid image that can be decoded/rendered progressively.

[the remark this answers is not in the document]
cconcolato: The grid images would not have the 'lsel' property because they are not coded images, they don't have layers. Only the referenced images have layers and can have 'lsel'. 

Now that the example is removed, do you think we should have a grid example?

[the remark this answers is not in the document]
wantehchang: I don't think we need a grid example. If we add the grid example, a good place to add it may be near this NOTE at line 212 because it is similar to what the NOTE describes:

`NOTE: When such a progressive decoding of the layers within an [=Operating Point=] is not desired or when an author wants to expose each layer as a specific item, multiple [=AV1 Image Items=] sharing the same [=AV1 Image Item Data=] can be created and associated with different 'lsel' properties, each with a different value of 'layer_id'.`

index.bs
```diff
@@ -176,19 +185,101 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
+
+[[!AV1]] delegates the selection of which [=Operating Point=] to process to the application. AVIF defines the [=OperatingPointSelectorProperty=] to control this selection. In the absence of an [=OperatingPointSelectorProperty=] associated with an [=AV1 Image Item=], the AVIF renderer is free to process any [=Operating Point=] present in the [=AV1 Image Item Data=]. In particular, when the [=AV1 Image Item=] is composed of a unique [=Operating Point=], the [=OperatingPointSelectorProperty=] should not be present. If an [=OperatingPointSelectorProperty=] is associated with an [=AV1 Image Item=], the <code>op_index</code> field indicates which [=Operating Point=] is expected to be processed for this item.
```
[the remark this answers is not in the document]
cconcolato: added

index.bs
cconcolato: I'm not sure we want to add a reference to a specific section, but I added an indication that it's in the "general output process" section.

cconcolato: Good point. My interpretation of the HEIF spec is that the use of `lsel` is codec-specific. We could ask to have that sentence clarified in HEIF.

index.bs
cconcolato: Good catch. Fixed now.

index.bs
cconcolato: I opened an issue on HEIF: https://github.com/MPEGGroup/FileFormat/issues/24

PR #125, opened 2021-01-06
add restriction on transformative properties in derivation chains

### LINKED ISSUE #124: Support for derivation chains
AVIF inherits the notion of image derivation (transformation, grids, overlays) from HEIF and the constraints on so-called "derivation chain" from MIAF (see section 7.3.11.1) which says:
```
A MIAF file may include any derived image defined in ISO/IEC 23008-12 with the constraints defined in the
following clauses.
A MIAF player shall process all derived images of the types identified in 7.3.7.
The maximum derivation chain is shown in the following list; the Identity, Grid and Overlay derivations are
optional. All derivation chains shall originate from one or more coded images. An identity derivation shall
not be derived immediately from another identity derivation. If derivations occur, they shall be in this
order:
• Mandatory Coded Image(s)
• Optional Identity derivation (7.3.11.2)
• Optional Grid (7.3.11.4)
• Optional Identity derivation (7.3.11.2)
• Optional Overlay (7.3.11.3)
• Optional Identity derivation (7.3.11.2)
```

This means you cannot create a grid of grids but you can create a grid of rotated images or a grid of cropped images. I created such image and it is currently not supported in libavif. Should we support it or should we restrict AVIF to only support transformation (rotation, mirroring, cropping) on the final image?

Attached image (zipped because GitHub does not support AVIF yet. Do we know how to request support for it?)
[clap_grid_2x2_2048x858.avif.zip](https://github.com/AOMediaCodec/av1-avif/files/5576330/clap_grid_2x2_2048x858.avif.zip)

cconcolato:
closes #124

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/125.html" title="Last updated on Jan 6, 2021, 6:03 PM UTC (75b7b68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/125/9459ef4...75b7b68.html" title="Last updated on Jan 6, 2021, 6:03 PM UTC (75b7b68)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -244,11 +244,14 @@ Additionally, if a file contains [=AV1 image sequences=] and the brand <code>[=a
 
 NOTE: As defined in [[!MIAF]], a file that is primarily an image sequence still has at least an image item. Hence, it can also declare brands for signaling the image item.
 
-  <h2 id="file-constraints">File constraints</h3>
+  <h2 id="file-constraints">General constraints</h3>
 
   The following constraints are common to files compliant with this specification:
   - The file shall be compliant with the [[!MIAF]] specification and list 'miaf' in the [=compatible_brands=] field of the [=FileTypeBox=].
   - The file shall list <code>'[=avif=]'</code> or <code>'[=avis=]'</code> in the [=compatible_brands=] field of the [=FileTypeBox=].
+  - If transformative properties are used in derivation chains (as defined in [[MIAF]]), they shall only be associated with items that are not referenced by another derived item. For example, if a file contains a grid item and its referenced coded image items, cropping, mirroring or rotation transformations are only permitted on the grid item itself.
+
+  NOTE: This constraints further restricts files compared to [[MIAF]].
 
   <h2 id="profiles">Profiles</h2>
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -244,11 +244,14 @@ Additionally, if a file contains [=AV1 image sequences=] and the brand <code>[=a
 
 NOTE: As defined in [[!MIAF]], a file that is primarily an image sequence still has at least an image item. Hence, it can also declare brands for signaling the image item.
 
-  <h2 id="file-constraints">File constraints</h3>
+  <h2 id="file-constraints">General constraints</h3>
 
   The following constraints are common to files compliant with this specification:
   - The file shall be compliant with the [[!MIAF]] specification and list 'miaf' in the [=compatible_brands=] field of the [=FileTypeBox=].
   - The file shall list <code>'[=avif=]'</code> or <code>'[=avis=]'</code> in the [=compatible_brands=] field of the [=FileTypeBox=].
+  - If transformative properties are used in derivation chains (as defined in [[MIAF]]), they shall only be associated with items that are not referenced by another derived item. For example, if a file contains a grid item and its referenced coded image items, cropping, mirroring or rotation transformations are only permitted on the grid item itself.
```
leo-barnes, line 252: The only use-case I can think of that is even remotely plausible is if you have a HW encoder that only supports a certain minimum frame size (say 512x512). If you then want to create a 'grid' with tiles that are 256x256 this might now not be possible since there is no way for you to crop the input tiles.

On the other hand, I know HEVC supports to handle stuff like this, so I'm guessing AV1 does as well.

leo-barnes (approved):
Looks good to me. Still feels slightly risky to restrict it, but given that libavif and others are not really implementing it anyway I can't really see how it can cause issues.

index.bs
cconcolato: I agree with the use case. Some times encoders are limited to produce multiples of power of 2, so cropping in grids could be needed...

leo-barnes: Per meeting, this should not be an issue since AV1 supports cropping if needed.

PR #135, opened 2021-03-16
Correct usage of level vs. level index

rzumer:
For consistency with the remainder of the document, this clarifies a single instance where "level" is used to refer to a `seq_level_idx` value which maps to the "maximum parameters" level.

***
<a href="https://pr-preview.s3.amazonaws.com/rzumer/av1-avif/pull/135.html" title="Last updated on Mar 16, 2021, 9:53 PM UTC (cc1fbf2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/135/7536f4b...rzumer:cc1fbf2.html" title="Last updated on Mar 16, 2021, 9:53 PM UTC (cc1fbf2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -359,7 +359,7 @@ NOTE: As defined in [[!MIAF]], a file that is primarily an image sequence still 
 
   It is possible for a file compliant to this [=AV1 Image File Format=] to not be able to declare an AVIF profile, if the corresponding AV1 encoding characteristics do not match any of the defined profiles.
 
-  NOTE: [[!AV1]] supports 3 bit depths: 8, 10 and 12 bits, and the maximum dimensions of a coded image is 65536x65536, when coded with level 31.
+  NOTE: [[!AV1]] supports 3 bit depths: 8, 10 and 12 bits, and the maximum dimensions of a coded image is 65536x65536, when `seq_level_idx` is set to 31 (maximum parameters level).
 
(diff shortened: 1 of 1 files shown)


leo-barnes (approved):
LGTM!

index.bs
```diff
@@ -359,7 +359,7 @@ NOTE: As defined in [[!MIAF]], a file that is primarily an image sequence still
 
   It is possible for a file compliant to this [=AV1 Image File Format=] to not be able to declare an AVIF profile, if the corresponding AV1 encoding characteristics do not match any of the defined profiles.
 
-  NOTE: [[!AV1]] supports 3 bit depths: 8, 10 and 12 bits, and the maximum dimensions of a coded image is 65536x65536, when coded with level 31.
+  NOTE: [[!AV1]] supports 3 bit depths: 8, 10 and 12 bits, and the maximum dimensions of a coded image is 65536x65536, when coded with level index 31 (maximum parameters).
```
cconcolato, line 362: Thank you. If we refer to the table in section A.3 of the AV1 spec, when `seq_level_idx` is 31, the 'level' column says "maximum parameters", so overall I see why you want to avoid using "level 31", but the term "level index" is never mentioned either in the AV1 specification. Could we just say:
> 65536x65536, when `seq_level_idx` is set to 31 (maximum parameters level).

rzumer: Thank you, I agree with your comment and have made the suggested change.

cconcolato:
I had to modify the PR and made a separate PR #137.

PR #143, opened 2021-04-01
Update a1lx property syntax and semantics

### LINKED ISSUE #139: Unused items in a1lx box
The a1lx box has a fixed array size of 4, but an image may have less spatial layers than that. The spec should specify what value to set in unused entries (preferred), or specify them as "don't care".

### LINKED ISSUE #140: Last layer_size entry in a1lx is redundant
The last layer size can simply be inferred to be the entire Item size.

Not only does this shrink layer_size to only 3 entries, but by not coding the largest layer size it makes it much more likely to be coded as a u16, saving up to 10 bytes in the best case.

cconcolato:
closes #139 
closes #140

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/143.html" title="Last updated on Apr 13, 2021, 7:47 PM UTC (5e21341)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/143/eb8b6ed...5e21341.html" title="Last updated on Apr 13, 2021, 7:47 PM UTC (5e21341)">Diff</a>


index.bs
```diff
@@ -267,21 +267,28 @@ If this property is associated with an item, the 'ispe' property documents the d
...
+    unsigned int(FieldLength)[3] layer_size;
 }
 ```
 
 <h6 id ="layered-image-indexing-property-semantics" class="no-toc">Semantics</h6>
 
-<dfn noexport>layer_size</dfn> indicates the number of bytes corresponding to each layer in the item payload. Values in the array are in the order of the layer [=spatial_id=].
+<dfn noexport>layer_size</dfn> indicates the number of bytes corresponding to each layer in the item payload, except for the last layer. Values are provided in increasing order of [=spatial_id=]. A value of zero means that all the layers except the last one have been documented and following values shall be 0. The number of non-zero values shall match the number of layers in the image minus one.
+
+NOTE: The size of the last layer can be determined by substracting the sum of the sizes of all layers indicated in this property from the entire item size.
```
leo-barnes, line 289: Typo: `substracting` -> `subtracting`

index.bs
```diff
@@ -267,21 +267,28 @@ If this property is associated with an item, the 'ispe' property documents the d
...
 
+<div class="example">A property indicating [X,0,0] is used for an image item composed of 2 layers. The size of the first layer is X and the size of the second layer is ItemSize - X. Note that the [=spatial_id=] for the first layer does not necessarily match the index in the array that provides the size. In other words, in this case the index giving value X is 0, but the corresponding [=spatial_id=] could be 0, 1 or 2. Similarly, a property indicating [X,Y,0] is used for an image is made of 3 layers.</div>
```
leo-barnes, line 291: 👍

leo-barnes (approved):
Looks good. Minor typo.

index.bs
cconcolato: done.

PR #201, opened 2022-11-14
Fix issue #177

### LINKED ISSUE #177: Invalid test files: Netflix/avis/alpha_video.avif `iref` value is incorrect
I believe the intent of https://github.com/AOMediaCodec/av1-avif/blob/master/testFiles/Netflix/avis/alpha_video.avif is to be a file with both 'avis' and 'avif' brand support, including an alpha channel associated with the primary image item.

The file contains both a primary image with `item_ID` 4:
Screen Shot 2021-12-04 at 3 08 20 PM

as well as an alpha image with `item_ID` 3:
Screen Shot 2021-12-04 at 3 08 48 PM

But the `iref` box which should associate the alpha `item_ID` (3) -> primary image `item_ID` (4) instead associates `item_ID` (3) -> `item_ID` (0)
Screen Shot 2021-12-04 at 3 09 39 PM

I believe the fix is simply changing the byte at offset 0x103 to 0x04 as in [alpha_video_fixed.avif.zip](https://github.com/AOMediaCodec/av1-avif/files/7654888/alpha_video_fixed.avif.zip).

leo-barnes:
Changed alpha_video.avif:
- Applied Jon's fix to set the correct iref
- Stripped out optional config OBUs from av1C
- Added explicit NCLX with 1,13,1,0
- Set track_in_movie to true in track

This closes #177


leo-barnes:
@cconcolato 
Do you want to sanity check the changed file? If not I'll merge this PR.

PR #223, opened 2024-08-13
Fix use of <dfn value> and some indenting.

### LINKED ISSUE #218: Use of `value` definition type for terms that are not CSS values
The spec uses `value` definition types for some of its definitions, e.g. [`av01`](https://aomediacodec.github.io/av1-avif/#valdef-av1-image-item-type-av01).

It's tempting to use the `value` type for such values but, despite the generic name, that type was defined for **CSS** values, see [Bikeshed's documentation](https://speced.github.io/bikeshed/#dfn-types). The use of `value` confuses tools that parse specs to extract semantic information, because that makes them think that the spec defines CSS terms.

There is no better definition type than `dfn` for now for such values. The possiblity to create one is being tracked in https://github.com/speced/spec-dfn-contract/issues/6

leo-barnes:
This fixes #218.

@tidoust, please take a look so that I understood the issue correctly.

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/223.html" title="Last updated on Sep 5, 2024, 8:25 AM UTC (b1e27f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/223/8ebeb5f...b1e27f9.html" title="Last updated on Sep 5, 2024, 8:25 AM UTC (b1e27f9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -435,4 +435,5 @@ The following additional constraints apply only to [=AV1 Image Sequences=]:
 <p>The media type <code>"image/avif"</code> is officially registered with IANA and available at: <a href="https://www.iana.org/assignments/media-types/image/avif">https://www.iana.org/assignments/media-types/image/avif</a>.</p>
 
 <h2 id="change-list">Changes since v1.1.0 release</h2>
+  - <a href="https://github.com/AOMediaCodec/av1-avif/pull/223">Stop using `dfn value` for definitions.</a>
 
--- a/index.bs
+++ b/index.bs
@@ -296,6 +296,7 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
 <div class="example">A property indicating [X,0,0] is used for an image item composed of 2 layers. The size of the first layer is X and the size of the second layer is ItemSize - X. Note that the [=spatial_id=] for the first layer does not necessarily match the index in the array that provides the size. In other words, in this case the index giving value X is 0, but the corresponding [=spatial_id=] could be 0, 1 or 2. Similarly, a property indicating [X,Y,0] is used for an image made of 3 layers.</div>
 
 <h2 id="image-sequences">Image Sequences</h2>
+
 <p>An <dfn export>AV1 Image Sequence</dfn> is defined as a set of AV1 [=Temporal Units=] stored in an [=AV1 track=] as defined in [[!AV1-ISOBMFF]] with the following constraints:
 
   - The track shall be a valid [=MIAF image sequence=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -168,17 +169,17 @@ When an item is of type <dfn value export for="AV1 Image Item Type">av01</dfn>,
...
-  Box Type:                 <dfn value export for="AV1 Item Configuration Property">av1C</dfn>
+  Box Type:                 <dfn export for="AV1ItemConfigurationProperty">av1C</dfn>
   Property type:            Descriptive item property
   Container:                ItemPropertyContainerBox
   Mandatory (per  item):    Yes, for an image item of type 'av01'
   Quantity:                 One for an image item of type 'av01'
 </pre>
 
-The syntax and semantics of the <dfn>AV1 Item Configuration Property</dfn> are identical to those of the [=AV1CodecConfigurationBox=] defined in [[!AV1-ISOBMFF]], with the following constraints:
+The syntax and semantics of the <dfn export>AV1ItemConfigurationProperty</dfn> are identical to those of the [=AV1CodecConfigurationBox=] defined in [[!AV1-ISOBMFF]], with the following constraints:
```
y-guyon, line 182: Should this be `[=AV1ItemConfigurationProperty=]` as it is already defined at line 175? What happens when there are multiple `<dfn>` for the same token? I looked at the [doc](https://speced.github.io/bikeshed/) but it was not clear.

index.bs
```diff
@@ -423,17 +421,18 @@ The following additional constraints apply only to [=AV1 Image Sequences=]:
...
+    <code>avis, msf1, miaf, MA1A</code>
 </div>
 
-  <h2 id="mime-registration">AVIF Media Type Registration</h2>
+<h2 id="mime-registration">AVIF Media Type Registration</h2>
 
 <p>The media type <code>"image/avif"</code> is officially registered with IANA and available at: <a href="https://www.iana.org/assignments/media-types/image/avif">https://www.iana.org/assignments/media-types/image/avif</a>.</p>
 
-  <h2 id="change-list">Changes since v1.1.0 release</h2>
+<h2 id="change-list">Changes since v1.1.0 release</h2>
```
y-guyon, line 437: Is this a list of changes in the source or in the generated document?

What does this PR do to the generated document?

y-guyon (approved):

index.bs
leo-barnes: I had to really dig into the BikeShed documentation and mess around with this to understand how it works. My understanding is that the definition on line 175 defines the term `AV1ItemConfigurationProperty/av1C` (i.e. `av1C` in the context of `AV1ItemConfigurationProperty`) as opposed to `av1C` as used in the sample entry in AV1-ISOBMFF. That is done to uniquely distinguish between the two uses.

The <dfn> on line 182 defines the term `AV1ItemConfigurationProperty` itself.

I tried changing line 182 to `[=AV1ItemConfigurationProperty=]`, but that causes BikeShed to complain about missing definitions for `AV1ItemConfigurationProperty`.

index.bs
leo-barnes: Ah. That is a good question. The generated document ends up _looking_ almost the same, but some of the defined terms in the document metadata are changed. So the Webref scraper will see different stuff for example.

The main noticeable difference for human readers is that `AV1 Item Configuration Property` is now `AV1ItemConfigurationProperty`.

leo-barnes: I guess we need to figure out what the policy is regarding which changes end up in this list.

cconcolato: You could list every PR and mark these kinds of PR as editorial.

leo-barnes:
Apparently Bikeshed requires indentation to be tabs or 4 spaces for some stuff. I'm converting the whole document to using 4 spaces everywhere so we're consistent.

leo-barnes:
Merging this so we can move forward with the other PR.s

leo-barnes:
Argh. Should have squashed as well. I expected GitHub to do that automatically.

index.bs
leo-barnes: > You could list every PR and mark these kinds of PR as editorial.

I ended up doing this.

PR #224, opened 2024-08-22
Sample transform

y-guyon:
Introduce Sample Transform Derived Image Items in AVIF.
Add the Definition, Syntax, Semantics and Constraints sections.

***
<a href="https://pr-preview.s3.amazonaws.com/y-guyon/av1-avif/pull/224.html" title="Last updated on Sep 17, 2024, 12:29 PM UTC (68400d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/224/2286f6c...y-guyon:68400d4.html" title="Last updated on Sep 17, 2024, 12:29 PM UTC (68400d4)">Diff</a>


index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+The output reconstructed image is made up of the output samples, whose values are each clamped to fit in the number of bits per sample as defined by the 'pixi' property of the reconstructed image item. The field <code>full_range_flag</code> of the 'colr' property of <code>colour_type</code> 'nclx' also defines a range of values to clamp to, as defined in [[!CICP]].
+
+<div class="example">A primary [=MIAF image item=] can be completed with another hidden [=MIAF image item=], both used as input image items to a [=Sample Transform Derived Image Item=], with an expression corresponding to ReconstructedSample = 256 × PrimarySample + HiddenSample. The primary [=MIAF image item=] and the [=Sample Transform Derived Image Item=] are both part of the same [=altr=] group. This will be perceived as a backward-compatible regular 8-bit image to readers not supporting [=Sample Transform Derived Image Items=], and can be decoded as a 16-bit image otherwise, making that pattern a bit-depth extension mechanism.</div>
```
leo-barnes, line 368: > A primary [=MIAF image item=] can be **completed**

Combined?

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+<dfn noexport>constant</dfn> is a literal signed value extracted from the stream with a precision of intermediate bit depth, pushed to the stack.
+
+<dfn noexport>input_image_item_index</dfn> is the 1-based index of the input image item whose sample is pushed to the stack. Shall be at most <code>[=reference_count=]</code>.
+
+<h5 id="sample-transform-constraints">Constraints</h5>
```
leo-barnes, line 607: Do we want to have a small section of "common recipes"? Or would that be better to have outside of the spec document itself? (Or potentially in an informative annex)

leo-barnes (commented):
Will review the changes in more detail soon, here are some immediate comments.

index.bs
y-guyon: I included the simplest bit depth extension example as a `<div class="example">`. I did not get into the details to keep it short and easy to understand.

The recipes are likely to evolve more often than the specification so I would lean toward another document. Does it have to be officially reviewed or can it be something like a wiki in this repository?

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+  - each have a 'pixi' property with version 1, an 'ispe' property and at least one 'colr' property;
+  - have the same number of channels and the same chroma subsampling (or lack thereof) as defined by the 'pixi' property;
+  - have the same dimensions as defined by the 'ispe' property;
+  - have the same colour type and space as defined by the 'colr' properties.
+
+For each output sample of the [=Sample Transform Derived Image Item=], a series of integer operations (an expression) is applied to operands. An operand is a constant or a sample from an input image item located at the same channel index and at the same spatial coordinates as the output sample.
+
+No colour space conversion, matrix coefficients, or transfer function shall be applied to the input samples. They are already in the same colour model as the output samples.
+
+Input samples shall have their field <code>channel_format</code> set to 0 (unsigned integer) or 3 (signed integer) as defined by the 'pixi' property. This constraint does not apply to output samples.
```
leo-barnes, line 364:
```suggestion
Input samples shall have their <code>channel_format</code> field set to 0 (unsigned integer) or 3 (signed integer) as defined by the 'pixi' property. This constraint does not apply to output samples.
```

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+The output reconstructed image is made up of the output samples, whose values are each clamped to fit in the number of bits per sample as defined by the 'pixi' property of the reconstructed image item. The field <code>full_range_flag</code> of the 'colr' property of <code>colour_type</code> 'nclx' also defines a range of values to clamp to, as defined in [[!CICP]].
```
leo-barnes, line 366:
```suggestion
The output reconstructed image is made up of the output samples, whose values are each clamped to fit in the number of bits per sample as defined by the 'pixi' property of the reconstructed image item. The <code>full_range_flag</code> field of the 'colr' property of <code>colour_type</code> 'nclx' also defines a range of values to clamp to, as defined in [[!CICP]].
```

leo-barnes: > whose values are each clamped to fit in the number of bits per sample

Do we want to make this a shall so that implementers don't miss it?

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+<h5 id="sample-transform-syntax">Syntax</h5>
+
+An expression is a series of tokens. A token is an operand or an operator. An operand can be a literal constant value or a sample value. A stack is used to keep track of the results of the subexpressions. An operator takes either one or two input operands. Each unary operator pops one value from the stack. Each binary operator pops two values from the stack, being the right operand then the left operand. Each token results in a value pushed to the stack. The single remaining value in the stack after evaluating the whole expression is the resulting output sample.
```
leo-barnes, line 372: Let's keep this as is for now, but it might be nice to make `operand` and `operator` defined terms that can be referred to.

leo-barnes, line 372: > The single remaining value in the stack after evaluating the whole expression is the resulting output sample.

We should probably try to word this as a `shall` so it's clear that this is a hard requirement on an expression.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+        unsigned int(8) token;
+        if (token == 0) {
+            // Push the 'constant' value to the stack.
+            signed int(1<<(bit_depth+3)) constant;
+        } else if (token == 1) {
+            // Push the sample value from the 'input_image_item_index'th
+            // input image item to the stack. At most 'reference_count'.
+            unsigned int(8) input_image_item_index; // 1-based
+        } else {
+            if (token == 2 || token == 3 || token == 11 || token == 12) {
```
leo-barnes, line 390: We might want to consider splitting the operators into two lists, where the binary operators have the top-most bit set or something like that. That way we would not need to change this `if` clause when new unary operators are added.

```suggestion
            if ((token & 0x80) == 0) { 
                // Unary operator. Pop an operand from the stack.
```

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+          </mtable>
+        </math>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+<dfn noexport>constant</dfn> is a literal signed value extracted from the stream with a precision of intermediate bit depth, pushed to the stack.
+
+<dfn noexport>input_image_item_index</dfn> is the 1-based index of the input image item whose sample is pushed to the stack. Shall be at most <code>[=reference_count=]</code>.
```
leo-barnes, line 605: Do we need to forbid `input_image_item_index` == 0?

index.bs
leo-barnes: > Does it have to be officially reviewed or can it be something like a wiki in this repository?

Let's keep it in the wiki!

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
 
+<h3 id="derived-images">Derived Image Items and Sequences</h3>
+
+<h4 id="sample-transform">Sample Transform Derived Image Item</h4>
+
+In these sections, a "sample" refers to the value of a pixel for a given channel.
+
+<h5 id="sample-transform-definition">Definition</h5>
+
+When a [=MIAF derived image item=] is of type <dfn value export for="Sample Transform Derived Image Item Type">sato</dfn>, it is called a <dfn>Sample Transform Derived Image Item</dfn>, and its reconstructed image is formed from a set of input image items, constants and operators.
```
leo-barnes, line 350:
```suggestion
When a [=MIAF derived image item=] is of type <dfn value export for="Sample Transform Derived Image Item Type">sato</dfn>, it is called a <dfn export>Sample Transform Derived Image Item</dfn>, and its reconstructed image is formed from a set of input image items, constants and operators.
```

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+            }
+            // Apply operator 'token' and push the result to the stack.
+        }
+    }
+}
+```
+
+<h5 id="sample-transform-semantics">Semantics</h5>
+
+<dfn noexport>version</dfn> shall be equal to 0.
```
leo-barnes, line 404: noexport is not strictly necessary since a pure `<dfn>` does not export by default. But I don't mind being explicit.

leo-barnes (approved):
Minor comments inline. Looks good to me!

index.bs
y-guyon: This paragraph is rather a description or an explanation.

The Constraints section already has a "shall":

> There shall be exactly one remaining element in the stack after evaluating the last token.

y-guyon: I doubt other sections or specifications will refer to `operand` and `operator`, especially without mentioning Sample Transforms.

And to avoid reserving these terms only for Sample Transforms they should probably be defined as "Sample Transform Derived Image Item operand/operator" which is very verbose.

index.bs
y-guyon: I did not follow closely what these tags were for. I mostly copy-pasted stuff.

index.bs
y-guyon: Suggestion:
- Move all binary operators to the end of the list
- Start binary operators with `token`=128

This avoids the need for two separate lists. The downside is the two ranges of reserved values.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+The input image items and the [=Sample Transform Derived Image Item=] shall:
+  - each have a 'pixi' property with version 1, an 'ispe' property and at least one 'colr' property;
+  - have the same number of channels and the same chroma subsampling (or lack thereof) as defined by the 'pixi' property;
+  - have the same dimensions as defined by the 'ispe' property;
+  - have the same colour type and space as defined by the 'colr' properties.
+
+For each output sample of the [=Sample Transform Derived Image Item=], a series of integer operations (an expression) is applied to operands. An operand is a constant or a sample from an input image item located at the same channel index and at the same spatial coordinates as the output sample.
+
+No colour space conversion, matrix coefficients, or transfer function shall be applied to the input samples. They are already in the same colour model as the output samples.
```
wantehchang, line 362: Nit: Add "characteristics" after "transfer". "Transfer characteristics" is the term used in CICP (H.273).

wantehchang, line 362: Nit: Use the American spelling of "color" in this document. Currently "colour" is only used in names defined in other standards, such as `ColourInformationBox`.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+  </thead>
+  <tbody>
+    <tr><td>0</td><td>8</td></tr>
+    <tr><td>1</td><td>16</td></tr>
+    <tr><td>2</td><td>32</td></tr>
+    <tr><td>3</td><td>64</td></tr>
+  </tbody>
+</table>
+
+The result of any computation under- or overflowing the intermediate bit depth is replaced by -2<sup><code>num_bits</code>-1</sup> and 2<sup><code>num_bits</code>-1</sup>-1, respectively. Encoder implementations should not create files leading to potential computation underflow or overflow. Decoder implementations shall check for computation underflow or overflow and adapt the results accordingly. Computations with operands of negative value use the two’s-complement representation.
```
wantehchang, line 428: Nit: Change "under-" to "underflowing"

wantehchang, line 428: Nit: Change "operands of negative value" to "operands of negative values" (plural "values") or "negative operands".

index.bs
```diff
@@ -435,3 +735,6 @@ A file containing a 'pict' track compliant with this profile is expected to list
 <p>The media type <code>"image/avif"</code> is officially registered with IANA and available at: <a href="https://www.iana.org/assignments/media-types/image/avif">https://www.iana.org/assignments/media-types/image/avif</a>.</p>
 
 <h2 id="change-list">Changes since v1.1.0 release</h2>
+
+  - <a href="https://github.com/AOMediaCodec/av1-avif/pull/224">Move section 4 "Auxiliary Image Items and Sequences" to section 4.1 under new section 4 "Other Image Items and Sequences".</a>
+  - <a href="https://github.com/AOMediaCodec/av1-avif/pull/224">Add section 4.2.1 "Sample Transform Derived Image Item" under new section 4.2 "Derived Image Items and Sequences".</a>
```
wantehchang, line 740: Nit/optional: It may be clearer to say:

> Add section 4.2 "Derived Image Items and Sequences" and section 4.2.1 "Sample Transform Derived Image Item".

index.bs
```diff
@@ -62,6 +62,15 @@ Metadata Order: This version, !*, *
   "status": "Enquiry",
   "publisher": "ISO/IEC",
   "isoNumber": "ISO/IEC DIS 23000-22"
+    },
```
wantehchang, line 65: (We should deal with this issue in a separate pull request.) Should we remove "DIS" from line 64 `"isoNumber": "ISO/IEC DIS 23000-22"` ?

wantehchang (approved):
LGTM. Thanks for converting the Sample Transforms proposal to this pull request.

I will review this more carefully after it is merged. I don't know how to view the HTML-rendered version of this file in the code review tool.

index.bs
wantehchang: It would be good to have the two bit depth extension recipies in some spec so they can be standardized, which will enable implementations to handle them as well-known special cases.

y-guyon:
> I don't know how to view the HTML-rendered version of this file in the code review tool.

There are two links in the first https://github.com/AOMediaCodec/av1-avif/pull/224#issue-2480821985: "Preview | Diff"  
They should be updated with each commit.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+      <td>2</td>
+      <td>Bitwise inclusive disjunction of the operands.</td>
+      <td><math><mi>L</mi><mo>∨</mo><mi>R</mi></math></td>
+    </tr>
+    <tr>
+      <td>134</td>
+      <td>xor operator</td>
+      <td>2</td>
+      <td>Bitwise exclusive disjunction of the operands.</td>
+      <td><math><mi>L</mi><mo>⊻</mo><mi>R</mi></math></td>
```
wantehchang, line 553: This may be a regional difference. I have seen the symbol ⊕ used for the xor operator.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+      <td>5</td>
+      <td>bsr operator</td>
+      <td>1</td>
+      <td>0-based index of the most significant set bit of the left operand if the left operand is strictly positive, zero otherwise.</td>
+      <td>
+        <math>
+          <mo>{</mo>
+          <mtable>
+            <mtr><mtd><mn>0</mn></mtd><mtd><mo>if</mo><mi>L</mi><mo>≤</mo><mn>0</mn></mtd></mtr>
+            <mtr><mtd>truncate<mo>(</mo><msub><mo>log</mo><mn>2</mn></msub><mi>L</mi><mo>)</mo></mtd><mtd>otherwise</mtd></mtr>
```
wantehchang, line 493: We may need to define the `truncate()` function.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+        unsigned int(8) token;
+        if (token == 0) {
+            // Push the 'constant' value to the stack.
+            signed int(1<<(bit_depth+3)) constant;
+        } else if (token == 1) {
+            // Push the sample value from the 'input_image_item_index'th
+            // input image item to the stack. At most 'reference_count'.
+            unsigned int(8) input_image_item_index; // 1-based
+        } else {
+            if ((token & 0x80) == 0) {
```
wantehchang, line 390: Nit/optional: We can change `(token & 0x80) == 0` to the equivalent condition `token < 128`.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+    <tr><td>2</td><td>32</td></tr>
+    <tr><td>3</td><td>64</td></tr>
+  </tbody>
+</table>
+
+The result of any computation underflowing or overflowing the intermediate bit depth is replaced by -2<sup><code>num_bits</code>-1</sup> and 2<sup><code>num_bits</code>-1</sup>-1, respectively. Encoder implementations should not create files leading to potential computation underflow or overflow. Decoder implementations shall check for computation underflow or overflow and adapt the results accordingly. Computations with operands of negative values use the two’s-complement representation.
+
+<dfn noexport>token_count</dfn> is the expected number of tokens to read.
+
+<dfn noexport>token</dfn> determines the type of the operand (constant or input image item sample) or the operator (how to transform one or two operands into the result). See Table 2. Other values than the ones in the table below are reserved.
```
wantehchang, line 432: It would be good to note that `token` values greater than 1 and less than 128 are unary operators, and `token` values greater than or equal to 128 are binary operators.

wantehchang (approved):
LGTM. It is fine to merge this pull request first. I will review it in preview form and can post my comments after the PR is merged.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
 
 NOTE: [[!AV1]] supports encoding either 3-component images (whose semantics are given by the <code>matrix_coefficients</code> element), or 1-component images (monochrome). When an image requires a different number of components, multiple auxiliary images may be used, each providing additional component(s), according to the semantics of their <code>aux_type</code> field. In such case, the maximum number of components is restricted by number of possible items in a file, coded on 16 or 32 bits.
 
+<h3 id="derived-images">Derived Image Items and Sequences</h3>
```
wantehchang, line 342: Delete "and Sequences" because there is no such thing as a derived image sequence.

index.bs
```diff
@@ -435,3 +735,6 @@ A file containing a 'pict' track compliant with this profile is expected to list
 <p>The media type <code>"image/avif"</code> is officially registered with IANA and available at: <a href="https://www.iana.org/assignments/media-types/image/avif">https://www.iana.org/assignments/media-types/image/avif</a>.</p>
 
 <h2 id="change-list">Changes since v1.1.0 release</h2>
+
+  - <a href="https://github.com/AOMediaCodec/av1-avif/pull/224">Move section 4 "Auxiliary Image Items and Sequences" to section 4.1 under new section 4 "Other Image Items and Sequences".</a>
+  - <a href="https://github.com/AOMediaCodec/av1-avif/pull/224">Add section 4.2 "Derived Image Items and Sequences" and section 4.2.1 "Sample Transform Derived Image Item".</a>
```
wantehchang, line 740: Change "Derived Image Items and Sequences" to "Derived Image Items".

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+In these sections, a "sample" refers to the value of a pixel for a given channel.
+
+<h5 id="sample-transform-definition">Definition</h5>
+
+When a [=MIAF derived image item=] is of type <dfn value export for="Sample Transform Derived Image Item Type">sato</dfn>, it is called a <dfn export>Sample Transform Derived Image Item</dfn>, and its reconstructed image is formed from a set of input image items, constants and operators.
+
+The input images are specified in, and in the same order as the <code>[=SingleItemTypeReferenceBox=]</code> or <code>[=SingleItemTypeReferenceBoxLarge=]</code> entries of type 'dimg' for this [=Sample Transform Derived Image Item=] within the <code>[=ItemReferenceBox=]</code>. In the <code>[=SingleItemTypeReferenceBox=]</code> or <code>[=SingleItemTypeReferenceBoxLarge=]</code> of type 'dimg', the value of the <code>[=from_item_ID=]</code> field identifies the [=Sample Transform Derived Image Item=], and the values of the <code>[=to_item_ID=]</code> field identify the input images. There are <code>[=reference_count=]</code> input image items as specified by the <code>[=ItemReferenceBox=]</code>.
```
wantehchang, line 352: Nonconstructive criticism: The first sentence of this paragraph is long and complicated. It is very hard to understand. I think it would help if we could break it into two simpler sentences.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+The input image items and the [=Sample Transform Derived Image Item=] shall:
+  - each have a 'pixi' property with version 1, an 'ispe' property and at least one 'colr' property;
```
wantehchang, line 355: IMPORTANT: Why does the 'pixi' property need to have version 1? Doesn't this break backward compatibility?

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+  - have the same number of channels and the same chroma subsampling (or lack thereof) as defined by the 'pixi' property;
+  - have the same dimensions as defined by the 'ispe' property;
+  - have the same color type and space as defined by the 'colr' properties.
+
+For each output sample of the [=Sample Transform Derived Image Item=], a series of integer operations (an expression) is applied to operands. An operand is a constant or a sample from an input image item located at the same channel index and at the same spatial coordinates as the output sample.
+
+No color space conversion, matrix coefficients, or transfer characteristics function shall be applied to the input samples. They are already in the same color model as the output samples.
```
wantehchang, line 362: Nit: Perhaps change "color model" to "color space"? The term "color model" or "colour model" does not appear in HEIF and CICP (H.273).

index.bs
wantehchang: I see PixelInformationProperty version 1 is designed to be backward compatible with version 0. But this requires the current readers to assume any version > 0 is backward compatible with version 0. Neither libavif nor libheif does that:

libavif/src/read.c:
```
static avifBool avifParsePixelInformationProperty(avifProperty * prop, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
{           
    BEGIN_STREAM(s, raw, rawLen, diag, "Box[pixi]");
    AVIF_CHECK(avifROStreamReadAndEnforceVersion(&s, 0));
```

libheif/libheif/box.cc
```
Error Box_pixi::parse(BitstreamRange& range)
{   
  parse_full_box_header(range);
      
  if (get_version() != 0) {
    return unsupported_version_error("pixi");
  }
```

wantehchang, line 355: IMPORTANT: If Sample Transforms can be used with alpha auxiliary image items, the requirement to have at least one 'colr' property conflicts with the following recommendation in the AVIF spec:

> For [AV1 Alpha Image Item](https://aomediacodec.github.io/av1-avif/v1.1.0.html#av1-alpha-image-item) and [AV1 Alpha Image Sequence](https://aomediacodec.github.io/av1-avif/v1.1.0.html#av1-alpha-image-sequence), the ColourInformationBox should be omitted. If present, readers shall ignore it.

index.bs
```diff
@@ -318,6 +339,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+    <tr>
+      <td>130</td>
+      <td>product operator</td>
+      <td>2</td>
+      <td>Left operand multiplied by the right operand.</td>
+      <td><math><mi>L</mi><mo>×</mo><mi>R</mi></math></td>
+    </tr>
+    <tr>
+      <td>131</td>
+      <td>divide operator</td>
```
wantehchang, line 521: Nit: The names of the four common arithmetic operators have inconsistent styles. I suggest either of the following.

1. Name them by the output of the operator: sum, difference, product, **quotient**
2. Name them by the action (a verb) of the operator: **add**, **subtract**, **multiply**, divide

index.bs
y-guyon: > We may need to define the `truncate()` function.

Not sure where. The table contains text such as "The result is truncated toward zero (integer division)" so I do not think defining `truncate()` is necessary, but I agree it would be nice to have. Is it worth creating a definition section though?

index.bs
y-guyon: I am open to suggestions.

y-guyon:
Rebased.

index.bs
wantehchang: Yes, it is common to have a section that defines the mathmetical symbols and functions used in the spec. If `truncate()` is the only function we need to define, it is not necessary to add a definition section.

I found that there is a `trunc()` function in the Standard C Library:
https://en.cppreference.com/w/c/numeric/math/trunc

So I think it is well known that `truncate() `means truncating towards 0, i.e., discarding the fractional part after the decimal point. Let's ignore this issue.

index.bs
wantehchang: There are two reasons why the first sentence is hard to understand. The first is that it describes a complex set of boxes. Since I am not familiar with that set of boxes, I am not qualified to suggest changes. The second reason is the "are specified in, and in the same order as" part. It crams two notions into the sentence. Although it shortens the sentence, it makes the sentence even more complex.

We can address the second problem by splitting the first sentence into two sentences. One describes the complex set of boxes, and the other describes the order of the input images.

index.bs
y-guyon: I removed that constraint. I think the PR works without it because input items must "have the same color type and space as defined by the 'colr' properties (or lack thereof)".

y-guyon: I removed all references to the pixi version. PTAL.

wantehchang: You are right. "(or lack thereof)" covers the alpha auxiliary image items.

wantehchang: I will take a look. Thanks!

index.bs
```diff
@@ -319,6 +340,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
 
+<h3 id="derived-images">Derived Image Items</h3>
+
+<h4 id="sample-transform">Sample Transform Derived Image Item</h4>
+
+In these sections, a "sample" refers to the value of a pixel for a given channel.
+
+<h5 id="sample-transform-definition">Definition</h5>
+
+When a [=MIAF derived image item=] is of type <dfn value export for="Sample Transform Derived Image Item Type">sato</dfn>, it is called a <dfn export>Sample Transform Derived Image Item</dfn>, and its reconstructed image is formed from a set of input image items, constants and operators.
```
leo-barnes, line 351: Is there anything that restricts the input images to be MIAF image items? I don't _think_ it looks like that (unless that is implied by MIAF derived image items) which I think is a good thing.

I could think of cases where you might want the base image to be a MIAF/AVIF item but some of the other inputs be uncompressed or use some item property not allowed by MIAF (like `iscl`).

wantehchang (approved):
Yannis: I reviewed the latest version. It looks good. Thanks!

All the comments in this round are minor suggestions to help make the text read better.

I do have a few high-level comments/questions that require discussion. I will present them in the Image Container SG meeting tomorrow.

index.bs
```diff
@@ -319,6 +340,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+<h5 id="sample-transform-syntax">Syntax</h5>
+
+An expression is a series of tokens. A token is an operand or an operator. An operand can be a literal constant value or a sample value. A stack is used to keep track of the results of the subexpressions. An operator takes either one or two input operands. Each unary operator pops one value from the stack. Each binary operator pops two values from the stack, being the right operand then the left operand. Each token results in a value pushed to the stack. The single remaining value in the stack after evaluating the whole expression is the resulting output sample.
```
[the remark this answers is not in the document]
wantehchang: Adrian suggests the following, which is closer to your wording:

```
the first being the right operand and the second being the left operand
```

index.bs
```diff
@@ -319,6 +340,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+            signed int(1<<(bit_depth+3)) constant;
+        } else if (token == 1) {
+            // Push the sample value from the 'input_image_item_index'th
+            // input image item to the stack. At most 'reference_count'.
+            unsigned int(8) input_image_item_index; // 1-based
+        } else {
+            if (token < 128) {
+                // Unary operator. Pop an operand from the stack.
+            } else {
+                // Binary operator. Pop a right operand
```
wantehchang, line 392: Nit: Change "a" to "the" before "right operand" and "left operand"

"the" sounds better here, because these are not arbitrary left and right operands; they are the operands for the current operator.

index.bs
```diff
@@ -171,7 +190,7 @@ When an item is of type <dfn export for="AV1 Image Item Type">av01</dfn>, it is
 <h4 id="av1-item-configuration-property">AV1 Item Configuration Property</h4>
 
 <pre class="def">
-    Box Type:                 <dfn export for="AV1ItemConfigurationProperty">av1C</dfn>
+    Box Type:                 <dfn export property>av1C</dfn>
```
leo-barnes, line 193: Was this intentional? Won't this cause a conflict with AV1-ISOBMFF which should already be defining `av1C`.

farindk:
I've read the proposal. Interesting concept.

Did you consider to add operators for modifying the stack (duplicating the top stack element, exchanging the top two elements, rolling the top N entries)? 
I did some Postscript programming in the past and those operations were frequently used. See here: https://webmail.ugrad.math.ubc.ca/Flat/stack.html

Are there any examples for typical use cases? I'm especially wondering what a potential use for the 'bsr' operator could be.

leo-barnes:
@farindk 
> I've read the proposal. Interesting concept.
> 
> Did you consider to add operators for modifying the stack (duplicating the top stack element, exchanging the top two elements, rolling the top N entries)? I did some Postscript programming in the past and those operations were frequently used. See here: https://webmail.ugrad.math.ubc.ca/Flat/stack.html
> 
> Are there any examples for typical use cases? I'm especially wondering what a potential use for the 'bsr' operator could be.

The main driver for this was to allow for bit depth extension. But while investigating that, we found that we would most likely want to have at least two recipes:
1. Lossless bit depth extension where the MSB are coded as one (lossless) image and the LSB as a second (also lossless).
2. Lossy bit depth extension where the MSB are coded as one lossy image and the LSB are coded lossily as a residual compared to the decoded lossy MSB image. For this case you may potentially have overlap between the MSB and LSB so that the residual image help counteract compression losses.

While going down that path we realized that it wouldn't actually be much harder to make it more generic and support "arbitrary" arithmetic on images, and that this might be very helpful with other features or experiments in the future. @y-guyon did an experiment where you could simulate floating point via `sato` for example.

`sato` could also be used to do residual coding with codecs that don't support it and similar.

leo-barnes:
Regarding rearranging the stack, I think we came to the conclusion that you should be able to arrange your tokens in such a way that you don't need to rearrange the stack while decoding.

If you have some specific example in mind that would be good to have though.

farindk:
The first association that I had when reading the proposal was that `sato` might be used for non-destructive image layer blending (https://printtechnologies.org/standards/files/pdf-reference-1.6-addendum-blend-modes.pdf). But to do that efficiently, I missed:
- some kind of `if`, probably implemented as a safe 'conditional move' (cmov) instruction,
- cross-component access (to be able to compute R/(R+G+B), for example). This would also enable custom color-space conversions, or we could blend images according to an aux image or alpha channel (generalization of `iovl`). This could be implemented as a second `sample` token that takes the image reference ID and a component index,
- stack manipulation to make it more efficient if there are common subexpressions. In many cases, you cannot eliminate common subexpressions by reordering the equation. One example with 'if's:
image
from (http://www.simplefilter.de/en/basics/mixmods.html) uses $2A$ and $2A-1$ many times. Also squaring values could simply be written as (`X dup mul`), where 'X' can even be a more complex expression.
- Another idea is whether there could be a simple enum that lists the most frequent equations. This would make it easier for an implementation to provide optimized code. Maybe as an optional information such that implementations that do not know the enum value can still use the token sequence as a fallback.

leo-barnes:
> The first association that I had when reading the proposal was that `sato` might be used for non-destructive image layer blending (https://printtechnologies.org/standards/files/pdf-reference-1.6-addendum-blend-modes.pdf). But to do that efficiently, I missed:
> 
> * some kind of `if`, probably implemented as a safe 'conditional move' (cmov) instruction,
> * cross-component access (to be able to compute R/(R+G+B), for example). This would also enable custom color-space conversions, or we could blend images according to an aux image or alpha channel (generalization of `iovl`). This could be implemented as a second `sample` token that takes the image reference ID and a component index,
> * stack manipulation to make it more efficient if there are common subexpressions. In many cases, you cannot eliminate common subexpressions by reordering the equation. One example with 'if's:

These are good suggestions. We have been considering various operators for a long time, but in the end we stuck with the ones we have now for the following reasons:
1. We only have two clearly defined use-cases as of right now. Lossless and lossy bit-depth extension. For those we basically  only need basic arithmetic operators.
2. We debated having more complex operators like log, exp and so on. But in order to use those in a reasonable way you would either need to start doing stuff in floating point or fixed point. Both of which increased the spec complexity by a lot. And given that we did not have clear use-cases that needed them the worry was that they way we added them would not actually end up being useful.
3. Note that all of the current operators apply equally to all components, **in the native colorspace of the content**. In other words, if your content is YCbCr, the operators apply in YCbCr. We considered adding operators that operated per channel, but it again explodes in spec complexity without us having a clear use-case that requires it. If you want to be able to do stuff in RGB you would need to specify operators for applying all the various CICP matrix coefficients for example. So lots of complexity.

In the interest of shipping a stable and usable starting point we decided to keep the operators simple for now, but have a versioning scheme that allows for extension in the future if new use-cases are identified.

> * Another idea is whether there could be a simple enum that lists the most frequent equations. This would make it easier for an implementation to provide optimized code. The enum could be optional information such that implementations that do not know the enum value can still use the token sequence as a fallback.

We have been debating this back and forth as well. The conclusion was that we want everything to be coded the same way so that you can be "guaranteed" that implementations will actually support all of it, but then have an annex of "commonly used" recipes that implementations might want to create an optimized path for.

wantehchang:
Having an enum for commonly-used recipies is convenient, but a downside is the extra work to maintain a registry of the enum's values.

Dirk: You asked what the `bsr` operator could be used for. `bsr` is log base 2 truncated to an integer. As Leo mentioned, earlier versions of this proposal had the exp and log operators with the base as an operand. I could not see how the exp and log operators, with their values truncated to integers, could be used in **meaningful** Sample Transforms, so I objected to including them. In the end the exp operator was removed and the log operator (with an arbitrary base) was replaced by log base 2, i.e., the `bsr` operator.

index.bs
y-guyon: As discussed, let's have an annex with these patterns.

index.bs
```diff
@@ -319,6 +340,285 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+<h5 id="sample-transform-syntax">Syntax</h5>
+
+An expression is a series of tokens. A token is an operand or an operator. An operand can be a literal constant value or a sample value. A stack is used to keep track of the results of the subexpressions. An operator takes either one or two input operands. Each unary operator pops one value from the stack. Each binary operator pops two values from the stack, being the right operand then the left operand. Each token results in a value pushed to the stack. The single remaining value in the stack after evaluating the whole expression is the resulting output sample.
+
+```c
+aligned(8) class SampleTransform {
+    unsigned int(2) version = 0;
+    unsigned int(4) reserved;
```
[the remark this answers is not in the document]
y-guyon: Not sure. HEIF uses two `reserved = 0;` and two `reserved;`. Only one of these four fields has a semantics entry with

> `reserved` shall be equal to 0 in files conforming to this version of this specification. The value of `reserved` shall be ignored by readers.

index.bs
y-guyon: Should it be `// Unary operator. Pop the operand from the stack.` then?

index.bs
y-guyon: This should export a property and AV1-ISOBMFF should export a regular dfn I guess? Otherwise I cannot refer to the `av1C` property from this document.

leo-barnes:
Latest changes look good to me!
Since I just merged the `tmap` change that adds the same sections there are merge conflicts.

index.bs
y-guyon: Fixed by using `'[=AV1ItemConfigurationProperty/av1C=]'` instead. Also filed #244 to address that separately and reverted the changes unrelated to this PR.

y-guyon:
I will merge this PR and open new PRs for unresolved offline high-level comments. This should simplify the review process.

Feel free to open issues/PRs too if I missed anything here before merging.

index.bs
wantehchang: Yes, "the operand" refers to the operand for the current unary operator.

Issue #216, opened 2023-10-02
Shared coded data between image and sequence contradicts still_picture=1


y-guyon:
Running [libavif's ChangeSettingTest.AOM](https://github.com/AOMediaCodec/libavif/blob/12e066686892df1c8201cfb0d8d6c68ad248c872/tests/gtest/avifchangesettingtest.cc#L104-L110) with `-DAVIF_ENABLE_COMPLIANCE_WARDEN=ON` gives:

> Specification description: AVIF v1.0.0, 19 February 2019
> https://aomediacodec.github.io/av1-avif/
> [avif][Rule #3] Warning: [ItemId=1] still_picture flag set to 0
> [avif][Rule #4] Warning: [ItemId=1] reduced_still_picture_header flag set to 0

This matches [AV1-AVIF v1.1.0](https://aomediacodec.github.io/av1-avif/v1.1.0.html#image-item) (also v1.0.0):

> If the AV1 Image Item Data consists of a single frame (i.e. when using a single layer),
> - It should have its `still_picture` flag set to 1.
> - It should have its `reduced_still_picture_header` flag set to 1.

libavif currently generates a `meta` box and image items alongside tracks when encoding sequences. I think this is the expected behavior because of ISO 23008-12 (HEIF):
> Files containing an image sequence should also contain a file-level MetaBox with a primary item that is an image item as specified in Clause 6, for cases in which temporal presentation is either undesirable, or not possible (e.g. printing).  
> NOTE The primary item can share coded data with one of the intra-coded images in the sequence.

However the "shared coded data" will not have its `still_picture` and `reduced_still_picture_header` fields set to 0 because it is a sequence. So the note contradicts the other recommendations.

The recommendations on `still_picture` and `reduced_still_picture_header` were removed for Image Sequences in https://github.com/AOMediaCodec/av1-avif/pull/112. I suggest we do the same for Image Items.

leo-barnes:
@y-guyon 
Do we want to try to fix this for v1.2.0?

y-guyon:
I sent https://github.com/AOMediaCodec/av1-avif/pull/228 if we want to simplify the specification but I do not think the real issue is in AV1-AVIF.

[...]

PR #228, opened 2024-09-10
Replace recommendations regarding still picture flags in image items by a note

y-guyon:
***
<a href="https://pr-preview.s3.amazonaws.com/y-guyon/av1-avif/pull/228.html" title="Last updated on Sep 17, 2024, 11:39 AM UTC (f688785)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/228/3d2d175...y-guyon:f688785.html" title="Last updated on Sep 17, 2024, 11:39 AM UTC (f688785)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -134,8 +134,6 @@ url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf; spec: AV1; type: dfn;
 url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf; spec: AV1; type: dfn;
     text: mono_chrome
     text: color_range
-    text: still_picture
-    text: reduced_still_picture_header
     text: operating_points_cnt_minus_1
     text: choose_operating_point
     text: spatial_id
@@ -174,9 +172,6 @@ When an item is of type <dfn export for="AV1 Image Item Type">av01</dfn>, it is 
     - The content of an [=AV1 Image Item=] is called the <dfn export>AV1 Image Item Data</dfn> and shall obey the following constraints:
         - <assert>The [=AV1 Image Item Data=] shall be identical to the content of an [=AV1 Sample=] marked as 'sync', as defined in [[!AV1-ISOBMFF]].</assert>
         - <assert>The [=AV1 Image Item Data=] shall have exactly one [=Sequence Header OBU=].</assert>
-        - If the [=AV1 Image Item Data=] consists of a single frame (i.e. when using a single layer),
-            - <assert>It should have its <code>[=still_picture=]</code> flag set to 1.</assert>
-            - <assert>It should have its <code>[=reduced_still_picture_header=]</code> flag set to 1.</assert>
 
 <h3 id="image-item-properties">Image Item Properties</h3>
 
@@ -857,3 +852,4 @@ Other versions of the boxes shall not be used. "-" means that the box does not h
     - EDITORIAL: <a href="https://github.com/AOMediaCodec/av1-avif/pull/233">Update list of other item properties</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/243">Further clarify relationship between av1C, metadata OBUs and item properties</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/239">Add information on tmap, grpl and altr</a>
+    - <a href="https://github.com/AOMediaCodec/av1-avif/pull/228">Remove recommendations regarding still flags in image items.</a>
--- a/index.bs
+++ b/index.bs
@@ -130,8 +130,6 @@ url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf; spec: AV1; type: dfn;
     text: Temporal Unit
     text: Operating Point
     text: Intra Frame
-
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -438,4 +433,4 @@ A file containing a 'pict' track compliant with this profile is expected to list
 <h2 id="change-list">Changes since v1.1.0 release</h2>
     - EDITORIAL: <a href="https://github.com/AOMediaCodec/av1-avif/pull/223">Stop using `dfn value` for definitions.</a>
     - EDITORIAL: <a href="https://github.com/AOMediaCodec/av1-avif/pull/222">Add assert-ids in the spec for conformance file testing and ComplianceWarden</a>
-
+    - <a href="https://github.com/AOMediaCodec/av1-avif/pull/228">Remove recommendations regarding still flags in image items.</a>
```
wantehchang, line 436: Nit: add "picture" after "still"

It would also be good to say "still picture" in the commit message, if it does not exceed the recommended 50-character line length.

wantehchang (commented):
Should the commit message say "Fixes https://github.com/AOMediaCodec/av1-avif/issues/216"?

I think it is fine to remove the recommendation, even though I suggested how to rephrase the recommendation to make it precise in https://github.com/AOMediaCodec/av1-avif/issues/216#issuecomment-2342295829.

leo-barnes:
> I think it is fine to remove the recommendation, even though I suggested how to rephrase the recommendation to make it precise in [#216 (comment)](https://github.com/AOMediaCodec/av1-avif/issues/216#issuecomment-2342295829).

I'm also ok with removing the recommendation, but would also be ok with Wan-Teh's alternative wording. I guess the "fear" with removing it would be that encoders don't realize it's an option and end up writing larger than necessary files.

But on the other hand, if libavif sets the flags appropriately when creating files I don't think the spec necessarily needs to say anything more about it.

leo-barnes:
Per meeting: 
We may want to convert this into an informative note simply saying that the still_picture and reduced_still_picture_header flags may be used to reduce the header overhead when applicable.

This informs readers that it exists but avoids the complicated discussion on exactly when it can't be set.

index.bs
```diff
@@ -162,9 +160,6 @@ When an item is of type <dfn export for="AV1 Image Item Type">av01</dfn>, it is
     - The content of an [=AV1 Image Item=] is called the <dfn export>AV1 Image Item Data</dfn> and shall obey the following constraints:
         - <assert>The [=AV1 Image Item Data=] shall be identical to the content of an [=AV1 Sample=] marked as 'sync', as defined in [[!AV1-ISOBMFF]].</assert>
         - <assert>The [=AV1 Image Item Data=] shall have exactly one [=Sequence Header OBU=].</assert>
-        - If the [=AV1 Image Item Data=] consists of a single frame (i.e. when using a single layer),
-            - <assert>It should have its <code>[=still_picture=]</code> flag set to 1.</assert>
-            - <assert>It should have its <code>[=reduced_still_picture_header=]</code> flag set to 1.</assert>
```
leo-barnes, line 163: Per discussion, we could add something like the following here:

```
NOTE: 
File writers may want to set the <code>[=still_picture=]</code> and 
<code>[=reduced_still_picture_header=]</code> flags to 1 when possible in the [=AV1 Image Item Data=] 
so that header overhead is minimized.
```

wantehchang: Leo: Thanks for suggesting the text for the note. The note needs to mention `[=Sequence Header OBU=]` and make it clear that <code>[=still_picture=]</code> and  <code>[=reduced_still_picture_header=]</code> are syntax elements in the `[=Sequence Header OBU=]`. Another possible improvement is to clarify "header overhead" refers to the overhead in the AV1 sequence header and frame header. Perhaps "AV1 header overhead" would suffice.

leo-barnes: Sounds good to me!
@y-guyon, what do you think?

index.bs
y-guyon: > Nit: add "picture" after "still"

Done.

> Should the commit message say "Fixes https://github.com/AOMediaCodec/av1-avif/issues/216"?

I usually avoid that because it triggers a GitHub cross-reference at every rebase. I can insert it at squashing time if you believe it is worth it. Note that the default squash title contains the PR number.

> It would also be good to say "still picture" in the commit message, if it does not exceed the recommended 50-character line length.

I am planning on squashing as "Replace AV1 still_pic=1 recommendation by note".

index.bs
y-guyon: Done.

leo-barnes (approved):
LGTM!

index.bs
```diff
@@ -174,9 +174,8 @@ When an item is of type <dfn export for="AV1 Image Item Type">av01</dfn>, it is
...
+
+NOTE: File writers may want to set the <code>[=still_picture=]</code> and <code>[=reduced_still_picture_header=]</code> flags to 1 when possible in the [=Sequence Header OBU=] part of the [=AV1 Image Item Data=] so that AV1 header overhead is minimized.
```
wantehchang, line 178: Nit/optional: "may want to" sounds rather weak. I think "should" sounds better here. But I guess "should" would make this a recommendation, which is what we wanted to avoid in the first place?

PR #230, opened 2024-09-11
This closes #179

### LINKED ISSUE #179: "This version" link doesn't work as expected
The field "This version" link in the various versions of the spec all link to the latest version, rather then the expected version of the sped.

leo-barnes:
gh-pages bot did not seem to like "broken" symbolic links, so it removed latest-draft.html since it didn't point to anything before bikeshed is run.

I reworded and reordered the links slightly into something that I think makes more sense, but I'm happy to make changes if people think the previous order was better.


y-guyon (approved):
This PR looks good.

I am often confused by these links, I usually do not now if the "Previously approved version" is different from the "Latest approved version" because the current version is not clear. I do not have anything to suggest though.

leo-barnes:
> I am often confused by these links, I usually do not now if the "Previously approved version" is different from the "Latest approved version" because the current version is not clear. I do not have anything to suggest though.

I agree. It makes more sense when viewing the published versions though. When you do that, latest approved version is not necessarily the same as the published version you're looking at right now.

y-guyon:
> It makes more sense when viewing the published versions though. When you do that, latest approved version is not necessarily the same as the published version you're looking at right now.

Yes, but I often find myself wondering if anything changed between before and after clicking.

PR #232, opened 2024-09-11
Add restriction on 'clap' property

### LINKED ISSUE #188: Add restrictions on 'clap' item property
Per the last meeting, the consensus was that we should add a restriction like the following on 'clap' to minimise the risk of it being misused:
The origin of the 'clap' item property shall be anchored to 0,0 (top-left) _unless_ the full, un-cropped image item is included as a secondary item in an image collection.

This allows 'clap' to be used for dealing with HW limitations (since those typically only require padding along the right/bottom edge).
The exception also allows for more advanced camera features while making sure that users are not surprised about what is contained in the file.

leo-barnes:
This closes #188

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/232.html" title="Last updated on Sep 25, 2024, 7:00 AM UTC (01fcade)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/232/e82a232...01fcade.html" title="Last updated on Sep 25, 2024, 7:00 AM UTC (01fcade)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -236,6 +236,12 @@ NOTE: The dimensions indicated in the 'ispe' property might not match the values
 
 NOTE: The values of [=render_width_minus1=] and [=render_height_minus1=] possibly present in the AV1 bistream are not exposed in the AVIF container level.
 
+<h4 id="clean-aperture-property">Clean Aperture Property</h4>
+
+The semantics of the clean aperture property ('clap') as defined in [[!HEIF]] apply. In addition to the restrictions on transformative item property ordering specified in [[!MIAF]], the following restriction also applies:
+
+<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) unless the full, un-cropped image item is included as a secondary item in an image collection.</assert>
+
 <h4 id="other-item-property">Other Item Properties</h4>
 
 In addition to the Image Properties defined in this document, [=AV1 image items=] MAY also be associated with item properties defined in other specifications such as [[!HEIF]] and [[!MIAF]]. Examples of commonly used item properties are:
@@ -254,8 +260,6 @@ In addition to the Image Properties defined in this document, [=AV1 image items=
     - 'cmex'
 
 In general, it is recommended to use properties instead of [=Metadata OBUs=] in the [=AV1ItemConfigurationProperty=].
-
-NOTE: Although the clean aperture property ('clap') defined in [[!HEIF]] is applicable to AVIF, implementers of authoring tools should be aware of the possibility of unintended consequences since users may not realize image data outside the clap region is still in the file. A future revision of this specification may place normative restrictions on how clap can be used.
 
 <h3 id="layered-items">AV1 Layered Image Items</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -1171,3 +1171,4 @@ Other versions of the boxes shall not be used. "-" means that the box does not h
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/239">Add information on tmap, grpl and altr</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/228">Replace recommendations regarding still picture flags in image items by a note</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/224">Add section 4.2.2 "Sample Transform Derived Image Item"</a>
+    - <a href="https://github.com/AOMediaCodec/av1-avif/pull/232">Add restriction on usage of clap property</a>
--- a/index.bs
+++ b/index.bs
@@ -240,7 +240,7 @@ NOTE: The values of [=render_width_minus1=] and [=render_height_minus1=] possibl
 
 The semantics of the clean aperture property ('clap') as defined in [[!HEIF]] apply. In addition to the restrictions on transformative item property ordering specified in [[!MIAF]], the following restriction also applies:
 
-<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) of the input image unless the full, un-cropped image item is included as a secondary item in an image collection.</assert>
+<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) of the input image unless the full, un-cropped image item is included as a secondary non-hidden image item in an image collection.</assert>
 
 <h4 id="other-item-property">Other Item Properties</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -240,7 +240,7 @@ NOTE: The values of [=render_width_minus1=] and [=render_height_minus1=] possibl
 
 The semantics of the clean aperture property ('clap') as defined in [[!HEIF]] apply. In addition to the restrictions on transformative item property ordering specified in [[!MIAF]], the following restriction also applies:
 
-<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) unless the full, un-cropped image item is included as a secondary item in an image collection.</assert>
+<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) of the input image unless the full, un-cropped image item is included as a secondary item in an image collection.</assert>
 
 <h4 id="other-item-property">Other Item Properties</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -240,7 +240,7 @@ NOTE: The values of [=render_width_minus1=] and [=render_height_minus1=] possibl
 
 The semantics of the clean aperture property ('clap') as defined in [[!HEIF]] apply. In addition to the restrictions on transformative item property ordering specified in [[!MIAF]], the following restriction also applies:
 
-<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) of the input image unless the full, un-cropped image item is included as a secondary non-hidden image item in an image collection.</assert>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -204,14 +204,18 @@ NOTE: The dimensions indicated in the 'ispe' property might not match the values
 
 NOTE: The values of [=render_width_minus1=] and [=render_height_minus1=] possibly present in the AV1 bistream are not exposed in the AVIF container level.
 
+<h4 id="clean-aperture-property">Clean Aperture Property</h4>
+
+The semantics of the clean aperture property ('clap') as defined in [[!HEIF]] apply. In addition to the restrictions on transformative item property ordering specified in [[!MIAF]], the following restriction also applies:
+
+<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) unless the full, un-cropped image item is included as a secondary item in an image collection.</assert>
```
y-guyon, line 211:
```diff
The origin of the 'clap' item property shall be anchored to 0,0 (top-left) unless
the full, un-cropped image item is included as a secondary item in an image collection
+containing the item associated with that 'clap' property
```
?

y-guyon, line 211: It would also be convenient to explicitly say that a reader may ignore any 'clap' in all cases because the full image is always included in the same `altr` group (if I understood this PR correctly).

y-guyon, line 211: Alternative: `essential` shall be equal to 0 for a 'clap' item property. Readers shall treat a `clap` item property as non-essential even if `essential` is 1.

y-guyon (commented):
Good idea.

It should be part of the "Changes" section.

index.bs
leo-barnes: No, I don't think that's the same thing.

Say we have item 1 that is the base item.
We also have items 2, 3 and 4 which are `iden` items pointing to 1 that adds a `clap` to crop out specific people in the image.
As a last step, we create item 5, an `iovl` that composits items 2-4 together into some kind of collage.

I don't necessarily want to make items 1-4 alternatives of each other. And I don't want item 5 to ignore the `clap` on items 2-4.

y-guyon: What is the point of the image collection if the item is non-hidden?

leo-barnes: That would mean it's fine for a parser to always ignore the `clap`, even when the item is an input to some other derived item.

It would also mean it's fine for a parser to ignore a top-left anchored `clap`, which would lead to padding being shown for images that need the `clap` to deal with HW limitations.

y-guyon: Right. I was thinking about implementations that parse but ignore the `clap` values and that do not expose the secondary non-hidden item.

As an example there is the libavif API that decodes to an `avifImage` which has a `clap` field. That `clap` field is ignored by `avifdec` which always outputs a PNG containing the full, uncropped image. I guess this is not compliant but I can imagine others doing the same, just accessing the pixels and ignoring the multiple obscure metadata chunks.

So I would have liked to adapt the spec to just accept the current state of things, at least for items that do not have further transformations, but that is beyond this PR I guess.

leo-barnes: > What is the point of the image collection if the item is non-hidden?

Implementation specific. We simply want to make sure that if a non-zero origin `clap` is used, the full image has to also be viewable. As for why someone would want to create a file like that, it's up to the exact use-case. We had experimental features that played around with ideas like this. (HEIF is currently a bit vague on how an image collection that doesn't contain something like a `ster` or `altr` group is to be interpreted.)

leo-barnes: Ah. I see your point.

I think the real fix would be for `avifdec` to apply the crop when writing to PNG though. From what I understand from the conversation where Chrome resurrected support for `clap`, there was a real use-case where it was needed for some HW that had size limitations. The current behaviour from `avifdec` will decode and show the padding for those files which I don't think is the correct behaviour.

y-guyon: Thank you for the explanation. My question was rather:  
If the item is non-hidden, it is viewable (right?). So there is no need to mention an image collection.

y-guyon: Is Chrome applying `clap` now? https://github.com/AOMediaCodec/libavif/issues/1177 can be closed then.

y-guyon: I filed https://github.com/AOMediaCodec/libavif/issues/2427.

leo-barnes: > Is Chrome applying `clap` now? [AOMediaCodec/libavif#1177](https://github.com/AOMediaCodec/libavif/issues/1177) can be closed then.

I think so at least:
https://github.com/chromium/chromium/commit/3a13337543c1e7de6914f87cd6f02ab06751c572

leo-barnes: https://issues.chromium.org/issues/40054481

y-guyon, line 211: As said during the call, I would have liked stronger constraints such as dimensions but this is too complicated.

y-guyon (approved):

leo-barnes:
From meeting: Maybe we need to specify that the anchoring is top-left of **input** rather than output

index.bs
wantehchang: I confirm that Chrome requires the clean aperture to be anchored at (0, 0) (top left) of the coded input image, and does not impose other requirements on the clean aperture.

We can modify Chrome to require that less than 64 pixels are cropped from the width or height of the coded input image. This will support the use case of supporting encoders that require width or height to be a multiple of 64. After gaining experience of this constraint in Chrome, we can then add it to the AVIF spec.

leo-barnes: @wantehchang 
> We can modify Chrome to require that less than 64 pixels are cropped from the width or height of the coded input image. This will support encoders that require width or height to be a multiple of 64. After gaining experience with this constraint in Chrome, we can then add it to the AVIF spec.

I understand the want for having stronger constraints, but we've had stuff like that come back and bite us multiple times in the past. We currently know of no AV1 HW that requires larger alignment than 64, but for HEVC we have had HW that requires horizontal alignments of:
- 64
- 128
- 192
- 256

And vertical alignments of:
- 64
- 96
- 128

Assuming that all future AV1 HW will stick to 64x64 sounds like a dangerous assumption to me.

-----------------

Threat:
The main thing we want to protect against is some kind of editing service that promises to do lossless cropping of your edit and that simply applies a `clap` to the file **without the user knowing that the full image is still present**. 

Valid use cases we want to be able to support:
- Dealing with HW limitations
- Useful tool for more advanced capture concepts

From my memory of the discussion, everyone could agree on two mitigations that would more or less completely mitigate the threat while still supporting the valid use-cases:
- Limit the `clap` to be anchored to the top-left corner. (This makes it useless as a general purpose cropping tool)

unless
- The full un-cropped image is also present and visible in the file. (This ensures that if used for generic cropping, the user is made aware that the content is still present.)

leo-barnes: > So there is no need to mention an image collection.

Ah. I see your point. I guess we should be covered if we say that it needs to be a secondary "**non-hidden** image item" since image item means it's an image collection.

What I guess I'm trying to say is that I don't think it should be enough that the non-cropped image is part of an image sequence since that frame might have such a short duration that it's not really visible.

index.bs
```diff
@@ -204,14 +204,18 @@ NOTE: The dimensions indicated in the 'ispe' property might not match the values
 
 NOTE: The values of [=render_width_minus1=] and [=render_height_minus1=] possibly present in the AV1 bistream are not exposed in the AVIF container level.
 
+<h4 id="clean-aperture-property">Clean Aperture Property</h4>
+
+The semantics of the clean aperture property ('clap') as defined in [[!HEIF]] apply. In addition to the restrictions on transformative item property ordering specified in [[!MIAF]], the following restriction also applies:
+
+<assert>The origin of the 'clap' item property shall be anchored to 0,0 (top-left) of the input image unless the full, un-cropped image item is included as a secondary non-hidden image item.</assert>
```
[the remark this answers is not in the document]
y-guyon: Good point.

From HEIF:

> A hidden image item has (flags & 1) equal to 1 in its ItemInfoEntry. 

So that it is easy to check.  
The presence of a `clap` property and whether it covers the full image should be straightforward too.

However an image item can be hidden or not, and cropped or not, but these are properties tied to the item, so I do not see how to have a cropped version and a non-cropped version of the same image item. Is it why collections were mentioned at first? Can one crop a collection? I guess it could be done through `iden` for example but checking if the full image is available through derived image items sounds a bit of work.  
Should it be "two image items referring to the same coded image bytes"?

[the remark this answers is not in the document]
leo-barnes: > I guess it could be done through iden for example but checking if the full image is available through derived image items sounds a bit of work.
> Should it be "two image items referring to the same coded image bytes"?

The way I envision someone doing this is per NOTE 2 in HEIF section 6.6.2.1 Identity derivation:
```
NOTE 2 A derived image of type 'iden' can be used, for example, when it is desirable to
have both the original version of a coded image item and a cropped version of the same
coded image item (obtained through the 'clap' transformative item property) as non-
hidden image items.
```

There are other ways it could be done, but this feels like the most straight forward and is relatively easy for a parser to check.

If someone uses an alternative scheme to do it, I think that should be allowed by the spec. It might not necessarily be implemented in all parsers, but that is something that can later be implemented if it turns out to be useful.

PR #239, opened 2024-09-12
Add tmap and altr text

### LINKED ISSUE #237: We should mention 'tmap' in AVIF
We should say that 'tmap' is possible in AVIF.

leo-barnes:
This closes #237

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/239.html" title="Last updated on Sep 17, 2024, 10:22 AM UTC (a9d4bb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/239/24508d6...a9d4bb5.html" title="Last updated on Sep 17, 2024, 10:22 AM UTC (a9d4bb5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -355,7 +355,7 @@ A tonemap derived image item ('tmap') as defined in [[!HEIF]] may be used in an 
 
 <h2 id="groups">Entity groups</h2>
 
-The GroupsListBox ('<dfn for="AVIF">grpl</dfn>') defined in [[!ISOBMFF]] may be used to group multiple image items in a file together. The type of the group describes how the image items are related.
+The GroupsListBox ('<dfn for="AVIF">grpl</dfn>') defined in [[!ISOBMFF]] may be used to group multiple image items in a file together. The type of the group describes how the image items are related. <assert>Decoders should ignore groups of unknown type.</assert>
 
 <h3 id="altr-group">'altr' group</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -351,7 +351,7 @@ NOTE: [[!AV1]] supports encoding either 3-component images (whose semantics are 
 
 <h4 id="tonemap-derivation">Tonemap Derived Image Item</h4>
 
-A tonemap derived image item ('tmap') as defined in [[!HEIF]] may be used in an [=AVIF=] file. When present, the base image item and the 'tmap' image item should be grouped together by an '[=AVIF/altr=]' entity group as recommended in [[!HEIF]].
+A tonemap derived image item ('tmap') as defined in [[!HEIF]] may be used in an [=AVIF=] file. <assert>When present, the base image item and the 'tmap' image item should be grouped together by an '[=AVIF/altr=]' entity group as recommended in [[!HEIF]].</assert>
 
 <h2 id="groups">Entity groups</h2>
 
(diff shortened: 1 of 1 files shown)


leo-barnes:
@y-guyon 
FYI: I stole the new sections from the sample transform PR, so this will cause merging conflicts. :)

index.bs
```diff
@@ -342,6 +347,24 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
 
+<h3 id="derived-images">Derived Image Items</h3>
+
+<h4 id="tonemap-derivation">Tonemap Derived Image Item</h4>
+
+A tonemap derived image item ('tmap') as defined in [[!HEIF]] may be used in an [=AVIF=] file. <assert>When present, the base image item and the 'tmap' image item should be grouped together by an '[=AVIF/altr=]' entity group as recommended in [[!HEIF]].</assert>
+
+<h2 id="groups">Entity groups</h2>
+
+The GroupsListBox ('<dfn for="AVIF">grpl</dfn>') defined in [[!ISOBMFF]] may be used to group multiple image items in a file together. The type of the group describes how the image items are related.
```
leo-barnes, line 358: I was debating whether we want to have language here saying that AVIF parsers are expected to understand the groups listed below or something saying that other group types than 'altr' and 'ster' may be present. But in the end I couldn't think of concise language to say it and simply skipped it.

y-guyon (approved):

index.bs
```diff
@@ -342,6 +347,24 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
 
 NOTE: [[!AV1]] supports encoding either 3-component images (whose semantics are given by the <code>matrix_coefficients</code> element), or 1-component images (monochrome). When an image requires a different number of components, multiple auxiliary images may be used, each providing additional component(s), according to the semantics of their <code>aux_type</code> field. In such case, the maximum number of components is restricted by number of possible items in a file, coded on 16 or 32 bits.
 
+<h3 id="derived-images">Derived Image Items</h3>
+
+<h4 id="tonemap-derivation">Tonemap Derived Image Item</h4>
```
wantehchang, line 352: Nit: Tonemap => Tone map

It is spelled as two words in [ISO/IEC DIS 23008-12/DAmd 1](https://www.iso.org/standard/89044.html).

index.bs
```diff
@@ -342,6 +347,24 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+
+A tonemap derived image item ('tmap') as defined in [[!HEIF]] may be used in an [=AVIF=] file. <assert>When present, the base image item and the 'tmap' image item should be grouped together by an '[=AVIF/altr=]' entity group as recommended in [[!HEIF]].</assert>
```
wantehchang, line 354: Nit: tonemap => tone map

wantehchang (approved):
LGTM. I am not that familiar with tmap, so I requested a review from Maryla.

@maryla-uc It would be good if you could review this pull request for the AVIF spec. Thanks!

index.bs
maryla-uc: In HEIF it's spelled inconsistently as "tone map" or "tone-map" but more often as the latter (with the hyphen), at least in the version of the spec that I have

index.bs
maryla-uc: The group type is the type of the enclosed box, and I think it's a general rule in isobmff that unknown box types should be ignored, so there should be no need to say anything.
In my opinion it doesn't hurt to repeat it anyway (i.e. "Decoders should ignore unknown group types", which implies that types other than altr and ster may be present). But since it's not done anywhere else in this spec, it might be confusing to only say it here?

maryla-uc (approved):

index.bs
leo-barnes: Added `<assert>Decoders should ignore groups of unknown type.</assert>`

index.bs
leo-barnes: Fixed

wantehchang (approved):

PR #240, opened 2024-09-12
Adopt MIAF shared constraints

### LINKED ISSUE #169: AVIF omits to select shared constraints from MIAF
MIAF defines constraints (Section 8) that can be shared by profiles, but they have to be explicitly selected. Among them is 'self-containment', which is assumed in AVIF but not explicitly stated. The others are: "single-layer", "grid-limit", "single-track", "edit-lists", "matched-duration". AVIF should clarify which ones are applicable.

leo-barnes:
This closes #169

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/240.html" title="Last updated on Sep 25, 2024, 7:07 AM UTC (4215905)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/240/adc56d5...4215905.html" title="Last updated on Sep 25, 2024, 7:07 AM UTC (4215905)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -136,6 +136,12 @@ url: https://www.iso.org/standard/74417.html; spec: MIAF; type: dfn;
     text: MIAF image sequence
     text: MIAF auxiliary image item
     text: MIAF auxiliary image sequence
+    text: MIAF file
+    text: self-containment
+    text: grid-limit
+    text: single-track
+    text: edit-lists
+    text: matched-duration
 
 url: https://aomediacodec.github.io/av1-isobmff/; spec: AV1-ISOBMFF; type: dfn;
     text: AV1CodecConfigurationBox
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -407,6 +412,15 @@ This section defines the MIAF AV1 Baseline profile of [[!HEIF]], specifically fo
 
 If the brand <code>MA1B</code> is in the list of [=compatible_brands=] of the [=FileTypeBox=], the common constraints in the section [[#brands]] shall apply.
 
+The following shared conditions and requirements from [[!MIAF]] shall apply:
```
leo-barnes, line 415: This is how it's done in MIAF. You can create a generic MIAF file that does not adopt these constraints, but if you want to use the specific profiles you're promising that you're adhering to these restrictions.

This means we can create a generic AVIF file that does not need to adhere to these restrictions. But if you want to use a specific profile you're promising that the file is well-behaved.

y-guyon (approved):
Could be in "the common constraints in the section [[#brands]]" to avoid the repetition

leo-barnes:
> Could be in "the common constraints in the section [[#brands]]" to avoid the repetition

See the comment inline. The reason MIAF did it this way is so that you can create a generic non-profiled MIAF file as well as MIAF files under specific profiles.

y-guyon:
> See the comment inline. The reason MIAF did it this way is so that you can create a generic non-profiled MIAF file as well as MIAF files under specific profiles.

Oh okay, I did not connect the two things. It makes sense now.

PR #245, opened 2024-09-17
Add Sample Transform rationale

y-guyon:
Make it obvious that AVIF also supports 16-bit.

***
<a href="https://pr-preview.s3.amazonaws.com/y-guyon/av1-avif/pull/245.html" title="Last updated on Sep 18, 2024, 7:50 AM UTC (a5d301c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/245/70d14e6...y-guyon:a5d301c.html" title="Last updated on Sep 18, 2024, 7:50 AM UTC (a5d301c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -375,7 +375,7 @@ A tone map derived image item ('tmap') as defined in [[!HEIF]] may be used in an
 
 In these sections, a "sample" refers to the value of a pixel for a given channel.
 
-With the Sample Transform file construction, pixels at the same position in multiple input image items can be aggregated into a single output pixel using basic mathematical operations. This makes it possible for [=AVIF=] to support 16 or more bits of precision per sample, while still offering backward compatibility through a regular 8 to 12-bit [=AV1 Image Item=] containing the most significant bits of each sample.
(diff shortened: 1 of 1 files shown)


leo-barnes (approved):

index.bs
```diff
@@ -375,6 +375,8 @@ A tone map derived image item ('tmap') as defined in [[!HEIF]] may be used in an
 
 In these sections, a "sample" refers to the value of a pixel for a given channel.
 
+With the Sample Transform file construction, pixels at the same position in multiple input image items can be aggregated into a single output pixel using basic mathematical operations. This makes it possible for [=AVIF=] to support 16 or more bits of precision per sample, while still offering backward compatibility through a regular 8 to 12-bit [=AV1 Image Item=] containing the most significant bits of each sample.
```
wantehchang, line 378: The term "file construction" is not used in the AVIF spec. Is "file construction" a term defined in HEIF?

Should this say "Sample Transform Derived Image Item" instead of "Sample Transform file construction"?

wantehchang, line 378: Nit: change "aggregated" to "combined"

The reason is that "aggregate" means "combine by adding" or "gather into a whole", so it may imply certain mathematical operations such as addition or concatenation. Using a less specific word like "combine" seems better.

wantehchang (approved):

index.bs
y-guyon: I avoided "Sample Transform Derived Image Item" because it is not yet defined. Besides that I do not mind.

PR #251, opened 2024-09-25
Clean up use of dfn and linking

### LINKED ISSUE #244: Inconsistent use of Bikeshed dfn and property references
There is no consistent pattern of using `dfn` or `property` as anchor definitions:

https://github.com/AOMediaCodec/av1-avif/blob/2286f6c1a642310f8d42523a1a784b0a8e7dfb8b/index.bs#L71-L148

There is no consistent pattern of referencing to anchors:

https://github.com/AOMediaCodec/av1-avif/blob/2286f6c1a642310f8d42523a1a784b0a8e7dfb8b/index.bs#L173

https://github.com/AOMediaCodec/av1-avif/blob/2286f6c1a642310f8d42523a1a784b0a8e7dfb8b/index.bs#L178

https://github.com/AOMediaCodec/av1-avif/blob/2286f6c1a642310f8d42523a1a784b0a8e7dfb8b/index.bs#L341

https://github.com/AOMediaCodec/av1-avif/blob/2286f6c1a642310f8d42523a1a784b0a8e7dfb8b/index.bs#L384

---

I suggest replacing all occurrences by:

- `type: dfn`+`[=token=]` for a **concept** (example "AV1 Image Item")
- `type: dfn`+`<code>token</code>` for a **struct** name, **field** name or **literal** value (examples `AV1ItemConfigurationProperty`, `still_picture`, `avif`)
- Optional: `type: property`+`'token'` strictly for an **image item property** (example `'colr'`)  
This is optional because I am not so sure Bikeshed properties were designed for something such as image item properties. I believe it is just a coincidence if the word "property" is used for both what Bikeshed was initially designed to support and the concept of an image item property.

Thoughts?

leo-barnes:
This closes #244

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/251.html" title="Last updated on Sep 25, 2024, 5:45 PM UTC (b9a7de9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/251/1796dae...b9a7de9.html" title="Last updated on Sep 25, 2024, 5:45 PM UTC (b9a7de9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -196,7 +196,7 @@ This specification reuses syntax and semantics used in [[!AV1-ISOBMFF]].
 
 When an item is of type <dfn export for="AV1 Image Item Type">av01</dfn>, it is called an <dfn export>AV1 Image Item</dfn>, and shall obey the following constraints:
     - <assert>The [=AV1 Image Item=] shall be a conformant [=MIAF image item=].</assert>
-    - <assert>The [=AV1 Image Item=] shall be associated with an [=AV1ItemConfigurationProperty=].</assert>
+    - <assert>The [=AV1 Image Item=] shall be associated with an <code>[=AV1ItemConfigurationProperty=]</code>.</assert>
     - The content of an [=AV1 Image Item=] is called the <dfn export>AV1 Image Item Data</dfn> and shall obey the following constraints:
         - <assert>The [=AV1 Image Item Data=] shall be identical to the content of an [=AV1 Sample=] marked as <code>'[=sync=]'</code>, as defined in [[!AV1-ISOBMFF]].</assert>
         - <assert>The [=AV1 Image Item Data=] shall have exactly one [=Sequence Header OBU=].</assert>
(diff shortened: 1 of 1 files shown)


leo-barnes:
In the end I did the following:
1. Anything that is a 4CC (for example all box types) gets referenced as `<code>'[=ispe=]'</code>`
2. Anything that is a struct field or variable or similar gets referenced as `<code>[=num_bits=]</code>`
3. Anything else simply gets referenced as is, without the `<code>`.

I also added a bunch of cross-linking in the 'sato' text so I could keep the <code> use without linking to a minimum.

leo-barnes:
I have removed <code> from any usage of the full box names in order to be consistent. We could revert this if you prefer it to be styled as it was previously, but then we should add it to all box names.

y-guyon (approved):
> Anything that is a struct field or variable or similar gets referenced as `<code>[=num_bits=]</code>`

I would also have put in `<code>` anything syntax-related, including struct names. For example `SingleItemTypeReferenceBox` is always rendered with a monospace font in ISOBMFF.

leo-barnes:
> > Anything that is a struct field or variable or similar gets referenced as `<code>[=num_bits=]</code>`
> 
> I would also have put in `<code>` anything syntax-related, including struct names. For example `SingleItemTypeReferenceBox` is always rendered with a monospace font in ISOBMFF.

I was debating how to handle that. If we want `<code>` around `SingleItemTypeReferenceBox` we then really need to add it around all box names. I agree that it's more consistent with ISOBMFF though. Let me try to add that.

index.bs
```diff
@@ -104,9 +104,10 @@ url: https://www.iso.org/standard/66067.html; spec: HEIF; type: property;
     text: image_height
     text: ster
     text: tmap
-
-url: https://www.iso.org/standard/66067.html; spec: HEIF; type: dfn;
     text: derived image item
+    text: aux_type
+    text: AuxiliaryTypeProperty
+    text: AuxiliaryTypeInfoBox
```
wantehchang, line 110: Optional: Reorder these two lines if you want to list these in alphabetical order. We seem to be appending to the end of the list most of the time.

wantehchang (approved):

PR #253, opened 2024-09-26
Add Appendix A on sato examples

### LINKED ISSUE #252: Add annex of `sato` examples
Per discussion

y-guyon:
Fixes #252

***
<a href="https://pr-preview.s3.amazonaws.com/y-guyon/av1-avif/pull/253.html" title="Last updated on Sep 30, 2024, 12:31 PM UTC (d987f5d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/253/1b5720e...y-guyon:d987f5d.html" title="Last updated on Sep 30, 2024, 12:31 PM UTC (d987f5d)">Diff</a>


index.bs
```diff
@@ -406,7 +407,7 @@ No color space conversion, matrix coefficients, or transfer characteristics func
...
 
-<div class="example">An 8-bit primary [=MIAF image item=] can be combined with another 8-bit hidden [=MIAF image item=], both used as input image items to a [=Sample Transform Derived Image Item=], with an expression corresponding to ReconstructedSample = 256 × PrimarySample + HiddenSample. The primary [=MIAF image item=] and the [=Sample Transform Derived Image Item=] are both part of the same <code>'[=AVIF/altr=]'</code> group. This will be perceived as a backward-compatible regular 8-bit image to readers not supporting [=Sample Transform Derived Image Items=], and can be decoded as a 16-bit image otherwise, making that pattern a bit-depth extension mechanism.</div>
+NOTE: <a href="#sato-examples">Appendix A</a> contains examples of Sample Transform Derived Image Item usage.
```
y-guyon, line 410: I was not able to make `[Appendix A](#sato-examples)` work despite being used in bikeshed doc:

https://github.com/speced/bikeshed/blob/828e63d35520ceba577f75b2d88ee92560d4bd58/docs/index.bs#L2631

`<a>` seems to give the same result anyway.

index.bs
```diff
@@ -659,7 +660,7 @@ The result of any computation underflowing or overflowing the intermediate bit d
 
 [=Sample Transform Derived Image Items=] use the postfix notation to evaluate the result of the whole expression for each reconstructed image item sample.
 
-  - <assert>The [=sato/tokens=] shall be evaluated in the order they are defined in the metadata (the <code>SampleTransform</code> structure) of the [=Sample Transform Derived Image Item=].</assert>
+  - <assert>The [=sato/tokens=] shall be evaluated in the order they are defined in the metadata (the <code><dfn export>SampleTransform</dfn></code> structure) of the [=Sample Transform Derived Image Item=].</assert>
```
y-guyon, line 663: Not sure about exporting this struct name.

index.bs
```diff
@@ -1195,6 +1196,90 @@ Other versions of the boxes shall not be used. "-" means that the box does not h
...
+
+Each output sample is equal to the sum of a sample of the first input image item shifted to the left by 4 and of a sample of the second input image item offset by -128. This can be viewed as a bit depth extension of the first input image item by the second input image item which contains the residuals to correct the precision loss of the first input image item.
+
+NOTE: If the first input image item is the [=primary image item=] and is enclosed in an <code>'[=AVIF/altr=]'</code> group with the derived image item, the first input image item is also a backward-compatible 12-bit regular coded image item that can be used by decoding contexts that do not support [=Sample Transform Derived Image Items=] or do not need extra precision.
+
+NOTE: The first input image item supports reasonable loss during encoding because the second input image item "overlaps" by 4 bits to correct it. The second input image item supports reasonable loss during encoding.
+
+NOTE: This pattern can be used for reconstructed bit depths beyond 16 by combining more than two input image items or with various input bit depth configurations and operations.
```
y-guyon, line 1285: For the title tree to make sense, there should be a section on floating point recipes here, but this is likely too involved.

index.bs
```diff
@@ -1195,6 +1196,90 @@ Other versions of the boxes shall not be used. "-" means that the box does not h
...
+
+This section contains demonstration recipes for extending base AVIF features with [=Sample Transform Derived Image Items=].
+
+<h3 id="sato-example-bit-depth-extension">Bit depth extension</h3>
+
+[=Sample Transform Derived Image Items=] allow for more than 12 bits per channel per sample by combining several [=AV1 image items=] in multiple ways.
+
+<h4 id="sato-example-suffix-bit-depth-extension">Suffix bit depth extension</h4>
+
+The following example describes how to leverage a [=Sample Transform Derived Image Item=] on top of a regular 8-bit [=MIAF image item=] to extend the decoded bit depth to 16 bits.
```
y-guyon, line 1214: It would be more accurate to put this whole section in a `<div class="example">` but I doubt it will work.

leo-barnes:
I don't love how appendices are rendered in Bikeshed, but there doesn't really seem to be any alternative. I've filed this issue to see if it's possible to get support for lettered annexes: https://github.com/speced/bikeshed/issues/2929

index.bs
leo-barnes: Yeah, I was debating whether to add it or not.

index.bs
```diff
@@ -1195,6 +1196,90 @@ Other versions of the boxes shall not be used. "-" means that the box does not h
...
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/228">Replace recommendations regarding still picture flags in image items by a note</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/224">Add section 4.2.2 "Sample Transform Derived Image Item"</a>
+    - <a href="https://github.com/AOMediaCodec/av1-avif/pull/253">Add Appendix A "Sample Transform Derived Image Item Examples"</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/232">Add restriction on usage of clap property</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/240">Adopt MIAF shared constraints</a>
     - EDITORIAL: <a href="https://github.com/AOMediaCodec/av1-avif/pull/251">Clean up usage of dfn and linking</a>
+
+<h2 id="sato-examples">Appendix A: Sample Transform Derived Image Item Examples</h2>
+
+This section contains demonstration recipes for extending base AVIF features with [=Sample Transform Derived Image Items=].
```
leo-barnes, line 1206: MPEG annexes typically specify whether an annex is informative or normative. Might be good to mention that this appendix is informative.

leo-barnes (approved):
I played around with various ways of trying to get a Letter.Number Annex tree, but it does not seem to be supported. I think this is the best we can do for now until Bikeshed implements something like https://github.com/speced/bikeshed/issues/2929

index.bs
y-guyon: Done

y-guyon:
Rebased.

index.bs
```diff
@@ -1194,7 +1195,91 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
+        - <code>[=sato/version=]</code>=0
+        - <code>[=sato/bit_depth=]</code>=2 (signed 32-bit <code>[=sato/constant=]</code>s, stack values and intermediate results)
+        - <code>[=sato/token_count=]</code>=7
+            - <code>[=sato/token=]</code>=0, <code>[=sato/constant=]</code>=16
+            - <code>[=sato/token=]</code>=1 (sample from 1<sup>st</sup> input image item)
+            - <code>[=sato/token=]</code>=130 (product)
+            - <code>[=sato/token=]</code>=2 (sample from 2<sup>nd</sup> input image item)
+            - <code>[=sato/token=]</code>=128 (sum)
+            - <code>[=sato/token=]</code>=0, <code>[=sato/constant=]</code>=-127
+            - <code>[=sato/token=]</code>=128 (sum)
```
wantehchang, line 1269: IMPORTANT: These two lines (1268-1269) mean adding the constant -127. It does not match the postfix and infix notations below, which subtract the constant 128.

index.bs
```diff
@@ -1194,7 +1195,91 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
+            - <math><msub><mi>sample</mi><mi>1</mi></msub><mo>≈</mo><mfrac><msub><mi>sample</mi><mi>original</mi></msub><msup><mn>2</mn><mn>4</mn></msup></mfrac></math>
+            - <math><msub><mi>sample</mi><mi>2</mi></msub><mo>≈</mo><msub><mi>sample</mi><mi>original</mi></msub><mo>-</mo><msup><mn>2</mn><mn>4</mn></msup><mo>×</mo><msub><mi>sample</mi><mi>1</mi></msub><mo>+</mo><msup><mn>2</mn><mn>7</mn></msup></math>
+            - <math><mn>0</mn><mo>≤</mo><msub><mi>sample</mi><mi>1</mi></msub><mo>&lt;</mo><msup><mn>2</mn><mn>12</mn></msup></math>
+            - <math><mn>0</mn><mo>≤</mo><msub><mi>sample</mi><mi>2</mi></msub><mo>&lt;</mo><msup><mn>2</mn><mn>8</mn></msup></math>
+            - <math><mn>0</mn><mo>≤</mo><msup><mn>2</mn><mn>4</mn></msup><mo>×</mo><msub><mi>sample</mi><mi>1</mi></msub><mo>+</mo><msub><mi>sample</mi><mi>2</mi></msub><mo>-</mo><msup><mn>2</mn><mn>7</mn></msup><mo>&lt;</mo><msup><mn>2</mn><mn>16</mn></msup></math>
```
wantehchang, line 1258: I remember I asked about this when I reviewed an earlier version of the Sample Transform proposal. If sample_1 and sample_2 are both equal to 0, they satisfy the first two inequalities (lines 1256 and 1257) but do not satisfy this inequality. I wonder if we need to clamp the expression in the middle of this inequality to the range [0, 2^16).

index.bs
wantehchang: Yes, the floating point recipes are complicated. I haven't reviewed them.

wantehchang (commented):
Yannis: Thank you for adding this appendix. I suggest some changes.

index.bs
leo-barnes: > I wonder if we need to clamp the expression in the middle of this inequality to the range [0, 2^16).

Good point. This will happen automatically though since the `sato` language says that the final result is clamped to the output bit-depth. But it might be good to call it out explicitly.

index.bs
leo-barnes: I think it's fine to leave them out for now. We may have other examples in the future that are not for bit depth extension. Having the title tree anticipate that makes it easier to add.

index.bs
```diff
@@ -697,12 +698,12 @@ The '<dfn for="AVIF">ster</dfn>' entity group as defined in [[!HEIF]] may be use
 The brand to identify [=AV1 image items=] is <dfn export for="AVIF Image brand">avif</dfn>.
 
 Files that indicate this brand in the <code>[=compatible_brands=]</code> field of the <code>[=FileTypeBox=]</code> shall comply with the following:
-    - <assert>The primary item shall be an [=AV1 Image Item=] or be a derived image that references directly or indirectly one or more items that all are [=AV1 Image Items=].</assert>
+    - <assert>The [=primary image item=] shall be an [=AV1 Image Item=] or be a derived image that references directly or indirectly one or more items that all are [=AV1 Image Items=].</assert>
```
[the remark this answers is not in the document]
y-guyon: I wanted to reference it with `[= =]`. Do you prefer:
- To keep as is
- To revert to `primary item`
- To use `[=primary item=]` and add `primary item` to the ISOBMFF `dfn`

index.bs
```diff
@@ -1194,7 +1195,91 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
+
+NOTE: The second input image item loses its meaning of least significant part if any of the most significant bits changes, so the first input image item has to be losslessly encoded. The second input image item supports reasonable loss during encoding.
+
+NOTE: This pattern can be used for reconstructed bit depths beyond 16 by combining more than two input image items or with various input bit depth configurations and operations.
+
+<h4 id="sato-example-residual-bit-depth-extension">Residual bit depth extension</h4>
+
+The following example describes how to leverage a [=Sample Transform Derived Image Item=] on top of a regular 12-bit [=MIAF image item=] to extend the decoded bit depth to 16 bits.<br>
+It differs from the <a href="sato-example-suffix-bit-depth-extension">Suffix bit depth extension example</a> by its slightly longer series of operations allowing its first input image item to be lossily encoded.
```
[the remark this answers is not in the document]
y-guyon: This is the name of a section, I thought section titles were copied verbatim when referenced.

index.bs
```diff
@@ -1194,7 +1195,91 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
+
+Consider the following:
+    - A [=MIAF image item=] being a lossily coded image item,<br>and its <code>'[=pixi=]'</code> property with <code>[=bits_per_channel=]</code>=12,
+    - Another [=MIAF image item=] being a lossily or losslessly coded image item with the same dimensions and number of samples as the first input image item,<br>and its <code>'[=pixi=]'</code> property with <code>[=bits_per_channel=]</code>=8,<br>with the following contraints:
```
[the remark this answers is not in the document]
y-guyon: Actually there are five constraints. I just did not manage to keep an indent layout that makes sense when rendered and that passes without warning in `bikeshed`.

The first part "For each sample position..." applies to each of the five following constraints.

index.bs
y-guyon: > I wonder if we need to clamp the expression in the middle of this inequality to the range [0, 2^16).

This is a constraint on the values of the input image item samples. Clamping would no longer make it a constraint.

If this is not clear enough that this is a constraint, could you suggest an alternative wording to make it explicit?

> Good point. This will happen automatically though since the `sato` language says that the final result is clamped to the output bit-depth. But it might be good to call it out explicitly.

We can add a note to remind about the clamping done at decoding.

index.bs
```diff
@@ -1194,7 +1195,91 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
+        - For each sample position in each plane,<br><math><msub><mi>sample</mi><mi>original</mi></msub></math> being the value of the 16-bit original sample at that position in that plane,<br><math><msub><mi>sample</mi><mi>1</mi></msub></math> being the value of the 12-bit sample of the first input image at that position in that plane,<br><math><msub><mi>sample</mi><mi>2</mi></msub></math> being the value of the sample of the second input image at that position in that plane,<br><math><mo>≈</mo></math> representing similarity within compression loss range,
```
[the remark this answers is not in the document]
y-guyon: It is not really a sentence, it is the premise of each of the five following constraints. Do you still think your comment applies?

Do you think repeating that premise at the beginning of each constraint instead of once before all would be better?

index.bs
y-guyon: Good catch, done.

index.bs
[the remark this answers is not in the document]
y-guyon: I encapsulated the premise in a hacky `<li style="list-style: none"><ul><li style="list-style: none">` `</li></ul></li>` to render it how I wanted to. Let me know if it is enough or if this is still unclear.

index.bs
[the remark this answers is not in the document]
y-guyon: I replaced the `<a>` by `[[#sato-example-suffix-bit-depth-extension]]` so there is a capital S introduced by bikeshed now.

index.bs
y-guyon: `[[#sato-examples]]` seems to work.

Issue #220, opened 2024-08-12
Multilayer text is confusing


leo-barnes:
We have this snippet in the spec (note bold text):
> In the absence of an [OperatingPointSelectorProperty](https://github.com/AOMediaCodec/av1-avif/issues/new#operatingpointselectorproperty) associated with an [AV1 Image Item](https://github.com/AOMediaCodec/av1-avif/issues/new#av1-image-item), the AVIF renderer is free to process any [Operating Point](https://aomediacodec.github.io/av1-spec/av1-spec.pdf) present in the [AV1 Image Item Data](https://github.com/AOMediaCodec/av1-avif/issues/new#av1-image-item-data). **In particular, when the [AV1 Image Item](https://github.com/AOMediaCodec/av1-avif/issues/new#av1-image-item) is composed of a unique [Operating Point](https://aomediacodec.github.io/av1-spec/av1-spec.pdf), the [OperatingPointSelectorProperty](https://github.com/AOMediaCodec/av1-avif/issues/new#operatingpointselectorproperty) should not be present**. If an [OperatingPointSelectorProperty](https://github.com/AOMediaCodec/av1-avif/issues/new#operatingpointselectorproperty) is associated with an [AV1 Image Item](https://github.com/AOMediaCodec/av1-avif/issues/new#av1-image-item), the op_index field indicates which [Operating Point](https://aomediacodec.github.io/av1-spec/av1-spec.pdf) is expected to be processed for this item.

I assume the text is trying to say that if the AV1 payload only consists of a single operating point, the a1op property should not be present, but I find the current text a bit confusing. How about something like this?

```diff
-In particular, when the AV1 Image Item is composed of a unique Operating Point, the OperatingPointSelectorProperty should not be present
+In particular, when the AV1 Image Item Data is composed of a single Operating Point, the OperatingPointSelectorProperty should not be present
```

leo-barnes:
@cconcolato @y-guyon @podborski 
Thoughts?

cconcolato:
Both are equivalent to me. The first one was referring to the logical structure (Image item), the proposed text is referring to the physical structure (image item data).

leo-barnes:
After thinking on it some more I think the existing text is actually more correct. Closing.

PR #256, opened 2024-10-09
Fix final review comments from Leo

leo-barnes:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/256.html" title="Last updated on Oct 16, 2024, 6:52 AM UTC (51a1b82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/256/e0e048e...51a1b82.html" title="Last updated on Oct 16, 2024, 6:52 AM UTC (51a1b82)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -79,118 +79,137 @@ Metadata Order: This version, !*, *
 
 <pre class="anchors">
 url: https://www.iso.org/standard/66067.html; spec: HEIF; type: dfn;
-    text: colr
-    text: mif1
-    text: msf1
-    text: pasp
-    text: pict
-    text: pixi
-    text: ispe
-    text: lsel
-    text: irot
-    text: imir
-    text: clap
-    text: cclv
-    text: clli
-    text: mdcv
-    text: reve
-    text: amve
-    text: ndwt
-    text: cmin
-    text: cmex
-    text: dimg
-    text: layer_id
-    text: image_width
-    text: image_height
-    text: ster
-    text: tmap
-    text: derived image item
     text: aux_type
+    text: auxC
     text: AuxiliaryTypeInfoBox
     text: AuxiliaryTypeProperty
+    text: auxl
     text: bits_per_channel
+    text: cdsc
+    text: cmex
+    text: cmin
+    text: derived image item
+    text: dimg
+    text: grid
+    text: image_height
+    text: image_width
+    text: imir
+    text: irot
+    text: ispe
+    text: layer_id
+    text: lsel
+    text: mif1
+    text: msf1
+    text: ndwt
+    text: pict
     text: PixelInformationProperty
+    text: pixi
+    text: prem
+    text: reve
+    text: ster
+    text: thmb
+    text: tmap
 
 url: https://www.iso.org/standard/68960.html; spec: ISOBMFF; type: dfn;
+    text: altr
+    text: amve
+    text: cclv
+    text: clap
+    text: clli
+    text: colour_type
+    text: ColourInformationBox
+    text: colr
     text: compatible_brands
+    text: ContentLightLevelBox
+    text: dinf
+    text: dref
     text: FileTypeBox
+    text: free
+    text: from_item_ID
+    text: ftyp
+    text: full_range_flag
+    text: GroupsListBox
+    text: grpl
+    text: hdlr
+    text: idat
+    text: iinf
+    text: iloc
+    text: infe
+    text: ipco
+    text: ipma
+    text: iprp
+    text: iref
+    text: ItemReferenceBox
     text: major_brand
+    text: MasteringDisplayColourVolumeBox
+    text: matrix_coefficients
+    text: mdat
+    text: mdcv
+    text: meta
+    text: nclx
+    text: pasp
+    text: pitm
+    text: reference_count
     text: SingleItemTypeReferenceBox
     text: SingleItemTypeReferenceBoxLarge
-    text: ItemReferenceBox
-    text: reference_count
-    text: from_item_ID
-    text: to_item_ID
-    text: nclx
-    text: sync
-    text: iloc
-    text: mdat
-    text: idat
-    text: altr
-    text: matrix_coefficients
-    text: full_range_flag
-    text: colour_type
-    text: meta
-    text: free
     text: skip
-    text: ItemPropertyContainerBox
-    text: MasteringDisplayColourVolumeBox
-    text: ContentLightLevelBox
+    text: sync
+    text: to_item_ID
 
 url: https://www.iso.org/standard/74417.html; spec: MIAF; type: dfn;
+    text: edit-lists
+    text: grid-limit
+    text: matched-duration
     text: miaf
-    text: primary image item
-    text: MIAF image item
-    text: MIAF image sequence
     text: MIAF auxiliary image item
     text: MIAF auxiliary image sequence
+    text: MIAF image item
+    text: MIAF image sequence
+    text: primary image item
     text: self-containment
-    text: grid-limit
     text: single-track
-    text: edit-lists
-    text: matched-duration
 
 url: https://aomediacodec.github.io/av1-isobmff/; spec: AV1-ISOBMFF; type: dfn;
-    text: AV1CodecConfigurationBox
     text: AV1 Sample
     text: AV1 Track
+    text: AV1CodecConfigurationBox
 
 url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf; spec: AV1; type: dfn;
     text: AV1 bitstream
     text: AV1 Frame
-    text: Sequence Header OBU
-    text: Metadata OBU
-    text: Temporal Unit
-    text: Operating Point
-    text: Intra Frame
-    text: mono_chrome
-    text: color_range
-    text: still_picture
-    text: reduced_still_picture_header
-    text: operating_points_cnt_minus_1
     text: choose_operating_point
-    text: spatial_id
-    text: seq_level_idx
-    text: render_width_minus1
-    text: render_height_minus1
-    text: UpscaledWidth
+    text: color_range
     text: FrameHeight
-    text: max_frame_width_minus1
+    text: Intra Frame
     text: max_frame_height_minus1
+    text: max_frame_width_minus1
+    text: Metadata OBU
+    text: mono_chrome
+    text: Operating Point
+    text: operating_points_cnt_minus_1
+    text: reduced_still_picture_header
+    text: render_height_minus1
+    text: render_width_minus1
+    text: seq_level_idx
+    text: Sequence Header OBU
+    text: spatial_id
+    text: still_picture
+    text: Temporal Unit
+    text: UpscaledWidth
 </pre>
 
 <h2 id="general">Scope</h2>
 
-[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn export>AV1 Image File Format</dfn> (AVIF) defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=] in a [[!HEIF]] file.
+[[!AV1]] defines the syntax and semantics of an [=AV1 bitstream=]. The <dfn export>AV1 Image File Format</dfn> (<dfn export>AVIF</dfn>) defined in this document supports the storage of a subset of the syntax and semantics of an [=AV1 bitstream=] in a [[!HEIF]] file.
 The [=AV1 Image File Format=] defines multiple profiles, which restrict the allowed syntax and semantics of the [=AV1 bitstream=] with the goal to improve interoperability, especially for hardware implementations.
 The profiles defined in this specification follow the conventions of the [[!MIAF]] specification.
-Images encoded with [[!AV1]] and not meeting the restrictions of the defined profiles may still be compliant to this [=AV1 Image File Format=] if they adhere to the general AVIF requirements.
+Images encoded with [[!AV1]] and not meeting the restrictions of the defined profiles may still be compliant to this [=AV1 Image File Format=] if they adhere to the general [=/AVIF=] requirements.
 
 The [=AV1 Image File Format=] supports High Dynamic Range (HDR) and Wide Color Gamut (WCG) images as well as Standard Dynamic Range (SDR). It supports monochrome images as well as multi-channel images with all the bit depths and color spaces specified in [[!AV1]], and other bit depths with [=Sample Transform Derived Image Items=]. The [=AV1 Image File Format=] also supports transparency (alpha) and other types of data such as depth maps through auxiliary [=AV1 bitstreams=].
 
 The [=AV1 Image File Format=] also supports multi-layer images as specified in [[!AV1]] to be stored both in image items and image sequences. The [=AV1 Image File Format=] supports progressive image decoding through layered images.
 
-An AVIF file is designed to be a conformant [[!HEIF]] file for both image items and image sequences. Specifically, this specification follows the recommendations given in "Annex I: Guidelines On Defining New Image Formats and Brands" of [[!HEIF]].
+An <dfn export>AVIF file</dfn> is designed to be a conformant [[!HEIF]] file for both image items and image sequences. Specifically, this specification follows the recommendations given in "Annex I: Guidelines On Defining New Image Formats and Brands" of [[!HEIF]].
 
 This specification reuses syntax and semantics used in [[!AV1-ISOBMFF]].
 
@@ -224,8 +243,8 @@ The syntax and semantics of the <dfn export>AV1ItemConfigurationProperty</dfn> a
     - <assert>[=Sequence Header OBUs=] should not be present in the <code>[=AV1ItemConfigurationProperty=]</code>.</assert>
     - <assert>If a [=Sequence Header OBU=] is present in the <code>[=AV1ItemConfigurationProperty=]</code>, it shall match the [=Sequence Header OBU=] in the [=AV1 Image Item Data=].</assert>
     - <assert>The values of the fields in the <code>[=AV1ItemConfigurationProperty=]</code> shall match those of the [=Sequence Header OBU=] in the [=AV1 Image Item Data=].</assert>
-    - <assert>The values of the bit depth and the number of channels derived from the <code>[=AV1ItemConfigurationProperty=]</code> shall match the <code>[=PixelInformationProperty=]</code> if present.</assert>
-    - <assert>[=Metadata OBUs=], if present, shall match the values given in other item properties</assert>, such as the <code>[=MasteringDisplayColourVolumeBox=]</code> or <code>[=ContentLightLevelBox=]</code>.
+    - <assert>The values of the bit depth and the number of channels derived from the <code>[=AV1ItemConfigurationProperty=]</code> shall match the <code>[=PixelInformationProperty=]</code> (<code>'[=pixi=]'</code>) if present.</assert>
+    - <assert>[=Metadata OBUs=], if present, shall match the values given in other item properties</assert>, such as the <code>[=MasteringDisplayColourVolumeBox=]</code> (<code>'[=mdcv=]'</code>) or <code>[=ContentLightLevelBox=]</code> (<code>'[=clli=]'</code>).
 
 <assert>This property should be marked as essential.</assert>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -274,11 +276,11 @@ In general, it is recommended to use properties instead of [=Metadata OBUs=] in
...
+[[!AV1]] delegates the selection of which [=Operating Point=] to process to the application, by means of a function called <code>choose_operating_point()</code>. [=/AVIF=] defines the <code>[=OperatingPointSelectorProperty=]</code> to control this selection. In the absence of an <code>[=OperatingPointSelectorProperty=]</code> associated with an [=AV1 Image Item=], the [=/AVIF=] renderer is free to process any [=Operating Point=] present in the [=AV1 Image Item Data=]. In particular, <assert>when the [=AV1 Image Item Data=] is composed of a unique [=Operating Point=], the <code>[=OperatingPointSelectorProperty=]</code> should not be present</assert>. If an <code>[=OperatingPointSelectorProperty=]</code> is associated with an [=AV1 Image Item=], the <code>[=op_index=]</code> field indicates which [=Operating Point=] is expected to be processed for this item.
```
leo-barnes, line 279: Regarding `AV1 Image Item` => `AV1 Image Item Data`, see issue #220.
In my mind, I see it like this:
1. The AV1 Image Item Data may contain one operating point or multiple operating points
2. The AV1 Image Item **always** consists of a unique operating point since the `a1op` defaults to selecting operating point 0 if not present.

So making statements about "when the AV1 Image Item is composed of a unique operating point" feels like something that is always true, but the text then says that if so the property should not be present.

Happy to revert if people disagree.

index.bs
```diff
@@ -371,23 +373,27 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
 
 <p>An <dfn export>AV1 Depth Image Item</dfn> (respectively an <dfn export>AV1 Depth Image Sequence</dfn>) is an [=AV1 Auxiliary Image Item=] (respectively an [=AV1 Auxiliary Image Sequence=]), and as defined in [[!MIAF]], with the <code>[=aux_type=]</code> field of the <code>[=AuxiliaryTypeProperty=]</code> (respectively <code>[=AuxiliaryTypeInfoBox=]</code>) set to <code>urn:mpeg:mpegB:cicp:systems:auxiliary:depth</code>.</p>
 
 NOTE: [[!AV1]] supports encoding either 3-component images (whose semantics are given by the <code>[=matrix_coefficients=]</code> element), or 1-component images (monochrome). When an image requires a different number of components, multiple auxiliary images may be used, each providing additional component(s), according to the semantics of their <code>[=aux_type=]</code> field. In such case, the maximum number of components is restricted by number of possible items in a file, coded on 16 or 32 bits.
 
 <h3 id="derived-images">Derived Image Items</h3>
 
+<h4 id="grid-derivation">Grid Derived Image Item</h4>
+
+A <dfn noexport>grid derived image item</dfn> (<code>'[=grid=]'</code>) as defined in [[!HEIF]] may be used in an [=AVIF file=].
```
leo-barnes, line 386: Added a small section on `grid` just so we can define it and refer to it in the rest of the text.

index.bs
```diff
@@ -371,23 +373,27 @@ NOTE: The size of the last layer can be determined by subtracting the sum of the
...
+A tone map derived image item (<code>'[=tmap=]'</code>) as defined in [[!HEIF]] may be used in an [=AVIF file=]. <assert>When present, the base image item and the <code>'[=tmap=]'</code> image item should be grouped together by an '[=AVIF/altr=]' entity group as recommended in [[!HEIF]].</assert>
 
 <h4 id="sample-transform">Sample Transform Derived Image Item</h4>
 
-In these sections, a "sample" refers to the value of a pixel for a given channel.
+With a [=Sample Transform Derived Image Item=], pixels at the same position in multiple input image items can be combined into a single output pixel using basic mathematical operations. This can be used to work around codec limitations or for storing alterations to an image as non-destructive residuals. With a [=Sample Transform Derived Image Item=] it possible for [=/AVIF=] to support 16 or more bits of precision per sample, while still offering backward compatibility through a regular 8 to 12-bit [=AV1 Image Item=] containing the most significant bits of each sample.
```
leo-barnes, line 394: Tried to make this description slightly more general so that it's clear that bit-depth extension is one use-case and others are possible.

index.bs
```diff
@@ -986,191 +992,191 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
         </tr>
         <tr>
             <td>&nbsp;</td>
             <td>dimg</td>
             <td>-</td>
-            <td>HEIF</td>
+            <td>[[!HEIF]]</td>
             <td>&nbsp;</td>
-            <td>Used when an image is derived from another image</td>
+            <td>Used when an image is [[#derived-images|derived from another image]]</td>
```
leo-barnes, line 1037: FYI: I noticed in the Bikeshed spec that you can provide alternative text for links by using `[=xxx|This is my alt text=]`

index.bs
```diff
@@ -1228,7 +1234,7 @@ Consider the following:
 
 This is equivalent to the following postfix notation (parentheses for clarity):
 
-<math><msub><mi>sample</mi><mi>output</mi></msub><mo>=</mo><mo>(</mo><mn>256</mn><mspace width="1ch"/><msub><mi>sample</mi><mn>1</mn></msub><mo>×</mo><mo>)</mo><msub><mi>sample</mi><mn>2</mn></msub><mo>+</mo></math>
+<math><msub><mi>sample</mi><mi>output</mi></msub><mo>=</mo><mo>(</mo><mn>256</mn><mspace width="0.5em"/><msub><mi>sample</mi><mn>1</mn></msub><mo>×</mo><mo>)</mo><mspace width="0.5em"/><msub><mi>sample</mi><mn>2</mn></msub><mo>+</mo></math>
```
leo-barnes, line 1237: `"1ch"` does not seem to work in Safari. The space simply disappears. `"0.5em"` seems to work in both Chrome and Safari and gives roughly the same space.

index.bs
```diff
@@ -469,11 +475,11 @@ aligned(8) class SampleTransform {
 
 The result of any computation underflowing or overflowing the intermediate bit depth is replaced by -2<sup><code>[=sato/num_bits=]</code>-1</sup> and 2<sup><code>[=sato/num_bits=]</code>-1</sup>-1, respectively. Encoder implementations should not create files leading to potential computation underflow or overflow. Decoder implementations shall check for computation underflow or overflow and clamp the results accordingly. Computations with [=sato/operands=] of negative values use the two’s-complement representation.
 
-<dfn noexport for="sato">token_count</dfn> is the expected number of [=sato/tokens=] to read.
+<dfn noexport for="sato">token_count</dfn> is the expected number of [=sato/tokens=] to read. The value of <code>[=sato/token_count=]</code> shall be greater than 0.
```
leo-barnes, line 478: I need to wrap this `shall` in `<assert>` if we want to keep it.

index.bs
leo-barnes: Should change `This can be used` to `This can for example be used`

index.bs
y-guyon: Same as above: why not `[=grid derived image item=]` directly, with a matching HEIF dfn entry?

index.bs
```diff
@@ -417,7 +423,7 @@ An expression is a series of [=sato/tokens=]. A [=sato/token=] is an [=sato/oper
 aligned(8) class SampleTransform {
     unsigned int(2) version = 0;
     unsigned int(4) reserved;
-    unsigned int(2) bit_depth; // Signed 8, 16, 32 or 64-bit.
+    unsigned int(2) bit_depth; // Enum signaling 8, 16, 32 or 64-bit.
```
y-guyon, line 426: "Signed" was intended, as in not unsigned

index.bs
```diff
@@ -469,11 +475,11 @@ aligned(8) class SampleTransform {
 
 The result of any computation underflowing or overflowing the intermediate bit depth is replaced by -2<sup><code>[=sato/num_bits=]</code>-1</sup> and 2<sup><code>[=sato/num_bits=]</code>-1</sup>-1, respectively. Encoder implementations should not create files leading to potential computation underflow or overflow. Decoder implementations shall check for computation underflow or overflow and clamp the results accordingly. Computations with [=sato/operands=] of negative values use the two’s-complement representation.
 
-<dfn noexport for="sato">token_count</dfn> is the expected number of [=sato/tokens=] to read.
+<dfn noexport for="sato">token_count</dfn> is the expected number of [=sato/tokens=] to read. <assert>The value of <code>[=sato/token_count=]</code> shall be greater than 0.</assert>
```
y-guyon, line 478: This is indirectly covered by the single remaining stack element constraint. I am fine with making it explicit here.

Note: We considered naming it `token_count_minus1` but it seemed inconsistent with other similar boxes.

index.bs
```diff
@@ -986,191 +992,191 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
...
         </tr>
         <tr>
             <td>a1op</td>
             <td>&nbsp;</td>
             <td>-</td>
-            <td>AVIF</td>
+            <td>[=/AVIF=]</td>
             <td>ipco</td>
-            <td>Used to configure rendering of a multiple operating points image</td>
+            <td>Used to configure rendering of an image containing multiple operating points</td>
```
y-guyon, line 1165: Based on your comment above, and assuming "image" here means "AV1 image item" and not "AV1 image item data", should it rather be:

> Used to configure the operating point to render an image with when there are multiple choices

y-guyon (approved):

index.bs
leo-barnes: See comment here https://github.com/AOMediaCodec/av1-avif/pull/256#discussion_r1793481437

index.bs
```diff
@@ -251,15 +253,15 @@ The semantics of the clean aperture property (<code>'[=clap=]'</code>) as define
 
 <h4 id="other-item-property">Other Item Properties</h4>
 
-In addition to the Image Properties defined in this document, [=AV1 image items=] MAY also be associated with item properties defined in other specifications such as [[!HEIF]] and [[!MIAF]]. Examples of commonly used item properties are:
-    - <code>'[=colr=]'</code>
-    - <code>'[=pixi=]'</code>
+In addition to the Image Properties defined in this document, [=AV1 image items=] may also be associated with item properties defined in other specifications such as [[!HEIF]] and [[!MIAF]]. Examples of commonly used item properties are:
+    - <code>'[=colr=]'</code> (<dfn noexport for="AVIF">ColourInformationBox</dfn>)
```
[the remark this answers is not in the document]
y-guyon: I feel it adds indirection. When one wants to look `ColourInformationBox` up, it will first lead to a place in AVIF, which itself redirects to HEIF. I get that this behavior is wanted to map `colr` to `ColourInformationBox`, but it also introduces confusion about which spec defines what.

> It's very hard to figure out how the various specs fit together.

I do not see an easy solution to this.

> It's unclear which things in HEIF and MIAF are actually supported/allowed in AVIF.

Solutions I can think of:
- Rely on the fact that everything that is not explicitly forbidden is allowed (~ current situation, could be written somewhere though)
- Succinctly list all boxes or features that are allowed (kind of an extended section 9. Box requirements) and declare all others as forbidden
- Create a new section for each supported box or feature, as was done for `tmap` and now `grid` etc. This feels like copy/pasting parts of ISOBMFF/HEIF/MIAF, but may be the clearest way for someone that just reads AV1-AVIF and not the inherited specs.

Example for `iden` which is allowed by AV1-AVIF but barely supported in implementations:
- Not explicitly forbidden hence allowed,
- Could be listed as `<code>[=iden=]</code>` and that's it,
- There could be a whole section about it.

Not sure which one is better than the others to be honest.

Anyway this seems beyond the topic of `AVIF/ColourInformationBox` and `AVIF/ColourInformationBox` barely answers that lack of clarity in my opinion.

I do not mind using both terms directly linked to 14496-12, especially if both are next to each other in the `dfn` list, but what about sticking to one of `colr` and `ColourInformationBox` if that is too much?

index.bs
```diff
@@ -1201,9 +1207,9 @@ The "Version(s)" column in the following table lists the version(s) of the boxes
     - EDITORIAL: <a href="https://github.com/AOMediaCodec/av1-avif/pull/251">Clean up usage of dfn and linking</a>
     - <a href="https://github.com/AOMediaCodec/av1-avif/pull/250">Clarify required versions of non-essential item properties</a>
```
y-guyon, line 1208: No `change-list` entry?

index.bs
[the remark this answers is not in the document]
leo-barnes: > Not sure which one is better than the others to be honest.

Yeah, that's what I was grappling with as well.

> I do not mind using both terms directly linked to 14496-12, especially if both are next to each other in the dfn list, but what about sticking to one of colr and ColourInformationBox if that is too much?

In order to improve consistency, how about this:
1. We switch all of the current usage of the full HEIF box names in AVIF to use the box types. That's mostly what we are using in any case, so it will only be a few changes. All of the box types that don't have explicit AVIF requirements or text links directly to the spec that defines it.
2. We add to the AVIF spec the boxes which are reasonably widely supported or something that we believe may end up being common in the foreseeable future. This means:
    1. For derived images we keep the sections for `grid` and `tmap` since they are widely supported. Since `iden` and `iovl` is only supported on Apple systems (I think?) it doesn't count as widely supported so we don't add sections for those. All use of `grid` and `tmap` inside of AVIF link to the AVIF sections which in turn links to HEIF.
    2. For item properties we add the properties that are widely supported or at least likely to be used to the list of boxes in `avif-required-boxes-additional`. 
    3. We either add common properties to the section "Other Item Properties" or completely remove that section and point it to `avif-required-boxes-additional`. I'm leaning towards the latter since it feels like we're duplicating stuff.

If we all agree on that, I guess the question comes down to which boxes to keeps in the list of boxes. 
- I added the `grpl`, `altr` and `ster` boxes since the first two are needed for `tmap` items and `ster` is pretty much the base requirement for stereo photos and we use them extensively for HEIC.
- I added all of the HDR item properties that have been added since they seem like they will likely be used now or in the future (and they are used in SlimHEIF).
- I also added the camera extrinsics and intrinsics since those are very important for stereo images and supported by libheif and the Apple platforms.

index.bs
leo-barnes: Yeah, it's weird. Maybe it doesn't work correctly in <math> tags but works in other CSS stuff?

index.bs
```diff
@@ -231,17 +233,17 @@ The syntax and semantics of the <dfn export>AV1ItemConfigurationProperty</dfn> a
...
 
     - In the absence of a <code>'[=lsel=]'</code> property associated with the item, or if it is present and its <code>[=layer_id=]</code> value is set to 0xFFFF:
-        - If no <code>[=OperatingPointSelectorProperty=]</code> is associated with the item, the <assert><code>'[=ispe=]'</code> shall document the dimensions of the last frame decoded when processing the operating point whose index is 0</assert>.
+        - If no <code>[=OperatingPointSelectorProperty=]</code> is associated with the item, the <assert><code>'[=ispe=]'</code> shall document the dimensions of the last frame decoded when processing the [=operating point=] whose index is 0</assert>.
```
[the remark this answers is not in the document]
leo-barnes: Right, but I'm not sure if the term really needs to be capitalized. At least this way they all link to the same thing. Happy to capitalize all mentions if you prefer though.

index.bs
```diff
@@ -221,8 +223,8 @@ The syntax and semantics of the <dfn export>AV1ItemConfigurationProperty</dfn> a
     - <assert>[=Sequence Header OBUs=] should not be present in the <code>[=AV1CodecConfigurationBox=]</code>.</assert>
     - <assert>If a [=Sequence Header OBU=] is present in the <code>[=AV1CodecConfigurationBox=]</code>, it shall match the [=Sequence Header OBU=] in the [=AV1 Image Item Data=].</assert>
     - <assert>The values of the fields in the <code>[=AV1CodecConfigurationBox=]</code> shall match those of the [=Sequence Header OBU=] in the [=AV1 Image Item Data=].</assert>
-    - <assert>The values of the bit depth and the number of channels derived from the <code>[=AV1CodecConfigurationBox=]</code> shall match the PixelInformationProperty if present.</assert>
-    - <assert>[=Metadata OBUs=], if present, shall match the values given in other item properties</assert>, such as the MasteringDisplayColourVolumeBox or ContentLightLevelBox.
+    - <assert>The values of the bit depth and the number of channels derived from the <code>[=AV1CodecConfigurationBox=]</code> shall match the <code>[=AVIF/PixelInformationProperty=]</code> if present.</assert>
```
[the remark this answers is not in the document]
leo-barnes: Same as the other discussion on defining locally vs linking directly. Marking this as resolved so we keep the discussion to a single thread.

index.bs
leo-barnes: Hmmm... Good catch. Let me think while I write.

Section 2.2.2 says:
- No `lsel` or `lsel == 0xFFFF`
    - `a1op` missing => `ispe` maps to last frame in operating point 0
    - `a1op` present => `ispe` maps to last frame in operating point specified by `a1op`
- `lsel != 0xFFFF`
    - `a1op` missing => `ispe` maps to frame selected by `lsel` (operating point not actually specified)
    - `a1op` present => `ispe` maps to frame selected by `lsel` **in operating point specified by `a1op`**

For the case of `lsel == 0xFFFF`, the renderer is free to choose intermediate frames for progressive decoding, but it's fully defined what is considered the final output image. If `a1op` is present, only the layers in that operating point are used. **If `a10p` is not present, I think the renderer is technically allowed to choose any operating point as long as the final frame is the last frame in operating point 0.**

For the case of `lsel` being present and selecting an actual layer, I guess **the renderer is technically allowed to select any operating point that contains the layer if there is no `a1op` property.**

For the case of `lsel` not being present, the AV1 data is not supposed to be multilayered. **I think this means you can't really have multiple operating points.**

-----------

To summarize, I think I now understand why @cconcolato considered both the old and the new text equivalent. You **can** have an image item that has an unspecified operating point. The final output layer is fully defined, but that layer may belong to multiple operating points, and the writer does not technically need to specify which to use.

Given this I think I'll revert this change.

index.bs
leo-barnes: Marking as resolved so we keep the discussion to a single thread.

index.bs
leo-barnes: Good point. Will change to say it's an enum signaling xxxx

index.bs
leo-barnes: Yeah, something like that would be better. Let me change it.

index.bs
leo-barnes: Will add

index.bs
leo-barnes: Sounds good. Keeping it explicitly required here

index.bs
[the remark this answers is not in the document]
y-guyon: > 1. We switch all of the current usage of the full HEIF box names in AVIF to use the box types

So `colr` everywhere and no mention of `ColourInformationBox`. This is fine with me, although I would have picked the latter, personally.

> 2. We add to the AVIF spec the boxes which are reasonably widely supported

Sounds good to me, as long as it is explicit enough that they are not _defined_ by AVIF.

> 2.iii. completely remove "Other Item Properties" section

Agreed.

> which boxes to keep

I trust you on that. I am not that knowledgeable on what is used out there.

index.bs
leo-barnes: Good question. That looks like it should be supported in Safari though. I pinged the WebKit folks to see if they know about it.

index.bs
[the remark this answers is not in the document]
y-guyon: Yes, I meant switching all, `colr` was just an example.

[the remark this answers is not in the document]
leo-barnes: Actually, looking through the doc, we use the box names consistently in some cases (`FileTypeBox`). Let me handle this as follows instead:
For the cases where we _mostly_ refer to the box by the 4CC but at some point use the name, make it look like this:
```
<code>[=MasteringDisplayColourVolumeBox=]</code> (<code>'[=mdcv=]'</code>)
```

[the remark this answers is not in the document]
leo-barnes: @y-guyon 
Let me know what you think. I've now made everything link directly to the base spec, _except in these places_:
1. `clap` in table of boxes links to `#clean-aperture-property` section
2. `grpl` in table of boxes links to `#groups` section
3. `altr` in table of boxes lins to `#altr-group` section
4. `ster` in table of boxes lins to `#ster-group` section

The places where we mention using `altr` groups now say something like this:
```
should be grouped together by an <code>'[=altr=]'</code> (see [[#altr-group]])
```

This way we link to the base spec _and_ add a link to the relevant section in AVIF.

index.bs
leo-barnes: Turns out it's an issue in Firefox as well. WebKit folks opened an issue for it:
https://bugs.webkit.org/show_bug.cgi?id=281232

index.bs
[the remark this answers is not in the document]
y-guyon: Sorry, I missed your replies. This looks good, thank you for the changes.

PR #264, opened 2024-10-18
Add major_brand along compats in Brands overview

y-guyon:
The ISOBMFF change allowing for the `major_brand` not to be repeated in the `compatible_brands` in `ftyp` was not yet published. This PR may stay open until so.

Whether it should be **and** or **or** in the following sentence will depend on the final wording of the published ISOBMFF change:

> Files that conform with these constraints should include the brand avif in the `major_brand` field **and**/**or** in the `compatible_brands` list.

***
<a href="https://pr-preview.s3.amazonaws.com/y-guyon/av1-avif/pull/264.html" title="Last updated on Oct 18, 2024, 11:56 AM UTC (56fa5c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/264/727501a...y-guyon:56fa5c2.html" title="Last updated on Oct 18, 2024, 11:56 AM UTC (56fa5c2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -703,7 +703,7 @@ The <code>'[=ster=]'</code> entity group as defined in [[!HEIF]] may be used to 
 
 <h3 id="brands-overview">Brands overview</h3>
 
-<p>As defined by [[!ISOBMFF]], the presence of a brand in the <code>[=compatible_brands=]</code> list in the <code>[=FileTypeBox=]</code> can be interpreted as the permission for those [=AV1 Image File Format=] readers/parsers and [=AV1 Image File Format=] renderers that only implement the features required by the brand, to process the corresponding file and only the parts (e.g. items or sequences) that comply with the brand.</p>
+<p>As defined by [[!ISOBMFF]], the presence of a brand in the <code>[=major_brand=]</code> field or in the <code>[=compatible_brands=]</code> list in the <code>[=FileTypeBox=]</code> can be interpreted as the permission for those [=AV1 Image File Format=] readers/parsers and [=AV1 Image File Format=] renderers that only implement the features required by the brand, to process the corresponding file and only the parts (e.g. items or sequences) that comply with the brand.</p>
 
 <p>An [=AV1 Image File Format=] file may conform to multiple brands. Similarly, an [=AV1 Image File Format=] reader/parser or [=AV1 Image File Format=] renderer may be capable of processing the features associated with one or more brands.</p>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -703,7 +703,7 @@ The <code>'[=ster=]'</code> entity group as defined in [[!HEIF]] may be used to
 
 <h3 id="brands-overview">Brands overview</h3>
 
-<p>As defined by [[!ISOBMFF]], the presence of a brand in the <code>[=compatible_brands=]</code> list in the <code>[=FileTypeBox=]</code> can be interpreted as the permission for those [=AV1 Image File Format=] readers/parsers and [=AV1 Image File Format=] renderers that only implement the features required by the brand, to process the corresponding file and only the parts (e.g. items or sequences) that comply with the brand.</p>
+<p>As defined by [[!ISOBMFF]], the presence of a brand in the <code>[=major_brand=]</code> field or in the <code>[=compatible_brands=]</code> list in the <code>[=FileTypeBox=]</code> can be interpreted as the permission for those [=AV1 Image File Format=] readers/parsers and [=AV1 Image File Format=] renderers that only implement the features required by the brand, to process the corresponding file and only the parts (e.g. items or sequences) that comply with the brand.</p>
```
leo-barnes, line 706: Do we want to simplify these changes and simply say something along the lines:
"As specified in ISOBMFF, the major brand shall be treated as if it is present among the compatible brands even if not explicitly specified."

Thoughts?

index.bs
```diff
@@ -712,32 +712,32 @@ The <code>'[=ster=]'</code> entity group as defined in [[!HEIF]] may be used to
...
+Additionally, the brand <dfn export for="AVIF Intra-only brand">avio</dfn> is defined. If the file indicates the brand <code>[=avio=]</code> in the <code>[=major_brand=]</code> field or in the <code>[=compatible_brands=]</code> list of the <code>[=FileTypeBox=]</code>, then <assert>the [=primary image item=] or all the items referenced by the [=primary image item=] shall be [=AV1 image items=] made only of [=Intra Frames=]</assert>. Conversely, <assert>if the previous constraint applies, the brand <code>[=avio=]</code> should be used in the in the <code>[=compatible_brands=]</code> list of the <code>[=FileTypeBox=]</code></assert>.
```
leo-barnes, line 721: It would be a bit weird to use `avio` as the major brand. Even though we're not explicitly suggesting anyone does so, just mentioning it as a possibility kind of increases the chances that people will do so.

index.bs
y-guyon: It would simplify this PR but not the spec itself in my opinion. It would make it slightly shorter though.

Maybe just say something such as "if `avif` is present in `ftyp`" or similar? This way it can implicitly be in `major_brand` and/or `compatible_brands` (but there could be confusion about `minor_version`).

y-guyon: Note that a similar wording already exists:

> Conversely, if such a track exists and the constraints of the brand [avio](https://github.com/AOMediaCodec/av1-avif/pull/264#avif-intra-only-brand-avio) on [AV1 image items](https://github.com/AOMediaCodec/av1-avif/pull/264#av1-image-item) are met, the brand should be used.

index.bs
y-guyon: Good point. I did not include `major_brand` in the second sentence though.

index.bs
leo-barnes: > Maybe just say something such as "if avif is present in ftyp" or similar? This way it can implicitly be in major_brand and/or compatible_brands (but there could be confusion about minor_version).

I like that. I don't _think_ anyone will confuse it with `minor_version`, but I guess we can add a sentence at the top to say that `minor_version` does not count as a brand or something like that.

leo-barnes: From meeting: new PR that tries to make it more general and just mentions having the brand in the `ftyp`.

y-guyon: Done in https://github.com/AOMediaCodec/av1-avif/pull/289.

leo-barnes:
I think this PR can be closed now, right?

Issue #269, opened 2024-10-20
-128, + vs. 128, - in the residual bit depth extension example


wantehchang:
In index.bs, we have:

```
            - <code>[=sato/token=]</code>=0, <code>[=sato/constant=]</code>=-128
            - <code>[=sato/token=]</code>=128 (sum)

This is equivalent to the following postfix notation (parentheses for clarity):

<math><msub><mi>sample</mi><mi>output</mi></msub><mo>=</mo><mo>(</mo><mo>(</mo><mn>16</mn><mspace width="0.5em"/><msub><mi>sample</mi><mn>1</mn></msub><mo>×</mo><mo>)</mo><mspace width="0.5em"/><msub><mi>sample</mi><mn>2</mn></msub><mo>+</mo><mo>)</mo><mspace width="0.5em"/><mn>128</mn><mo>-</mo></math>

This is equivalent to the following infix notation:

<math><msub><mi>sample</mi><mi>output</mi></msub><mo>=</mo><mn>16</mn><mo>×</mo><msub><mi>sample</mi><mn>1</mn></msub><mo>+</mo><msub><mi>sample</mi><mn>2</mn></msub><mo>-</mo><mn>128</mn></math>
```

In the `SampleTransform`, we have the constant `-128` and the operator `+` (sum). In the postfix and infix notations, we have the constant `128` and the operator `-`.  They do not match, even though they are mathematically equivalent.

y-guyon:
I remember seeing that and thought that was fine, but I agree it can be fixed for clarity.

PR #272, opened 2024-10-21
Use exact tokens as sato math expr example

y-guyon:
Fixes https://github.com/AOMediaCodec/av1-avif/issues/269.


wantehchang (approved):
LGTM. Thanks!

leo-barnes (approved):

PR #300, opened 2025-01-20
Make asserts between profiles unique

### LINKED ISSUE #299: Assert IDs for MIAF shared conditions in baseline and advanced profiles are the same
@rbouqueau pointed out that the assert IDs for the MIAF shared conditions in the AVIF profile are the same since they contain the same text.

Ideally we would make the asserts contain the section ID or something like that so that they can be validated as separate asserts.

It's unclear how to do this in a nice way though. Ideally we would not add any visible text to the conditions.

leo-barnes:
This closes #299

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/300.html" title="Last updated on Jan 20, 2025, 12:25 PM UTC (338bdb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/300/46332a5...338bdb1.html" title="Last updated on Jan 20, 2025, 12:25 PM UTC (338bdb1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -766,13 +766,13 @@ This section defines the MIAF AV1 Baseline profile of [[!HEIF]], specifically fo
 If the brand <code>'[=MA1B=]'</code> is in the <code>[=FileTypeBox=]</code>, the common constraints in the section [[#brands]] shall apply.
 
 The following shared conditions and requirements from [[!MIAF]] shall apply:
-    - <assert>[=self-containment=] (subclause 8.2)</assert>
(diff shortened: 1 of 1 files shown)


leo-barnes:
Checking with @cconcolato whether we can take editorial changes.

y-guyon (approved):

wantehchang (commented):
LGTM.

PR #304, opened 2025-01-24
Update HEIF,ISOBMFF,MIAF references to new version

wantehchang:
Update the HEIF, ISOBMFF, and MIAF references to the latest versions.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -41,29 +41,29 @@ Metadata Order: This version, !*, *
 
     "HEIF": {
         "id": "HEIF",
-        "href": "https://www.iso.org/standard/66067.html",
+        "href": "https://www.iso.org/standard/83650.html",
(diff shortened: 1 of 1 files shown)


leo-barnes (approved):

y-guyon (commented):
This is more than an editorial change, right? Can it be done during an ongoing review/publication process?

leo-barnes:
> This is more than an editorial change, right? Can it be done during an ongoing review/publication process?

Good point. Yeah, this might not count as purely editorial.

leo-barnes:
> > This is more than an editorial change, right? Can it be done during an ongoing review/publication process?
> 
> Good point. Yeah, this might not count as purely editorial.

On the other hand, we may want to tweak this anyway once the HEIF amendment containing `tmap` lands.

y-guyon (approved):

PR #314, opened 2025-08-20
Update HEIF to point to 3ed AMD1

### LINKED ISSUE #307: Update link to HEIF version to point to 3ed Amd1 FDAM
Per discussion, we're mentioning 'tmap' in v1.2.0. We should link to HEIF 3ed Amd1 FDAM since that is where it's added.

leo-barnes:
Closes #307

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/pull/314.html" title="Last updated on Aug 20, 2025, 4:56 PM UTC (0c6268f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/av1-avif/314/3ba0ec7...0c6268f.html" title="Last updated on Aug 20, 2025, 4:56 PM UTC (0c6268f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -41,11 +41,11 @@ Metadata Order: This version, !*, *
 
     "HEIF": {
         "id": "HEIF",
-        "href": "https://www.iso.org/standard/83650.html",
+        "href": "https://www.iso.org/standard/89044.html",
         "title": "Information technology — High efficiency coding and media delivery in heterogeneous environments — Part 12: Image File Format",
         "status": "International Standard",
         "publisher": "ISO/IEC",
-        "isoNumber":"ISO/IEC 23008-12:2022"
+        "isoNumber":"ISO/IEC 23008-12:2025/FDAmd 1"
     },
 
     "ISOBMFF": {


index.bs
```diff
@@ -41,11 +41,11 @@ Metadata Order: This version, !*, *
 
     "HEIF": {
         "id": "HEIF",
-        "href": "https://www.iso.org/standard/83650.html",
+        "href": "https://www.iso.org/standard/89044.html",
         "title": "Information technology — High efficiency coding and media delivery in heterogeneous environments — Part 12: Image File Format",
         "status": "International Standard",
         "publisher": "ISO/IEC",
-        "isoNumber":"ISO/IEC 23008-12:2022"
+        "isoNumber":"ISO/IEC 23008-12:2025/FDAmd 1"
```
wantehchang, line 48: Leo: This is amendment 1 of ISO/IEC 23008-12:2025. It seems that this should be a separate reference, say "HEIF FDAmd 1".

And then we can update the "HEIF" reference to ISO/IEC 23008-12:2025 at https://www.iso.org/standard/89035.html.

y-guyon: > I see no difference between referring to 23008-12:2022, 23008-12:2025 or 23008-12:2025/FDAmd1, they are all just various dated references to HEIF.

It is not a single branch though, is it?  
For example if one has to refer to both `tmap` and `mini`, you cannot refer to a single dated reference, as AMD2 does not contain AMD1 (or does it?).

y-guyon (approved):

index.bs
leo-barnes: > For example if one has to refer to both tmap and mini, you cannot refer to a single dated reference, as AMD2 does not contain AMD1 (or does it?).

I think it does. AMD2 refers to stuff in AMD1 as well as stuff in the fully integrated HEIF documents. That would not be possible if AMD1 was not treated as containing everything in the amendment itself plus the document it applies to.

wantehchang (approved):

index.bs
podborski: Leo is correct. AMD2 needs to refer to all the changes in AMD1. Otherwise one could have merge conflicts when new edition is created.

PR #315, opened 2025-08-26
Clarify 'sato' item requirements.

### LINKED ISSUE #309: Requirements on sato item and its input images
The current draft specification says:

> The input image items and the [Sample Transform Derived Image Item](https://aomediacodec.github.io/av1-avif/#sample-transform-derived-image-item) shall:
> -  each be associated with a [PixelInformationProperty](https://www.iso.org/standard/83650.html) and an '[ispe](https://www.iso.org/standard/83650.html)' property;
> -  have the same number of channels and the same chroma subsampling (or lack thereof) as defined by the [PixelInformationProperty](https://www.iso.org/standard/83650.html) and [AV1ItemConfigurationProperty](https://aomediacodec.github.io/av1-avif/#av1itemconfigurationproperty);
> -  have the same dimensions as defined by the '[ispe](https://www.iso.org/standard/83650.html)' property;
> - have the same color information as defined by the [ColourInformationBox](https://www.iso.org/standard/83102.html) properties (or lack thereof).

Does this mean that the sato item and its input image items must all have a colr property with the same content, OR must have no colr property associated with any of them? This is how I would naturally interpret "have the same color information as defined by the [ColourInformationBox](https://www.iso.org/standard/83102.html) properties (or lack thereof)" but this is not what libavif is currently outputting. In the images created in avif16bitttest.cc, neither the sato box nor the extra input items have any colr property (this may be a bug with libavif).

The same question applies to pixi/av1C. Especially if version 0 of pixi is used, then I believe the only way to signal chroma subsampling is through the av1C box. But then this makes it sound like the sato item should have an av1C box too.

leo-barnes:
This closes #309.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -412,8 +412,8 @@ The input images are specified in the <code>[=SingleItemTypeReferenceBox=]</code
 
 The input image items and the [=Sample Transform Derived Image Item=] shall:
   - each be associated with a <code>[=PixelInformationProperty=]</code> and an <code>'[=ispe=]'</code> property;
-  - have the same number of channels as defined by the<code>[=PixelInformationProperty=]</code> or <code>[=AV1ItemConfigurationProperty=]</code>;
-  - have the same chroma subsampling; this may be explicitly defined by one of the above properties for some input image items and/or the [=Sample Transform Derived Image Item=], and may be implicit for the other items (meaning no property defines chroma subsampling for these items);
+  - have the same number of channels as defined by the <code>[=PixelInformationProperty=]</code> or <code>[=AV1ItemConfigurationProperty=]</code>;
+  - have the same chroma subsampling; this may be explicitly defined by one of the above properties for some input image items or the [=Sample Transform Derived Image Item=], and may be implicit for the other items (meaning no property defines chroma subsampling for these items);
   - have the same dimensions as defined by the <code>'[=ispe=]'</code> property;
   - have the same color information as defined by the <code>[=ColourInformationBox=]</code> properties (or lack thereof).
 


index.bs
```diff
@@ -412,7 +412,8 @@ The input images are specified in the <code>[=SingleItemTypeReferenceBox=]</code
 
 The input image items and the [=Sample Transform Derived Image Item=] shall:
   - each be associated with a <code>[=PixelInformationProperty=]</code> and an <code>'[=ispe=]'</code> property;
-  - have the same number of channels and the same chroma subsampling (or lack thereof) as defined by the <code>[=PixelInformationProperty=]</code> and <code>[=AV1ItemConfigurationProperty=]</code>;
+  - have the same number of channels as defined by the<code>[=PixelInformationProperty=]</code> or <code>[=AV1ItemConfigurationProperty=]</code>;
```
y-guyon, line 415: Missing space after "the"

y-guyon (approved):

index.bs
```diff
@@ -412,7 +412,8 @@ The input images are specified in the <code>[=SingleItemTypeReferenceBox=]</code
...
+  - have the same chroma subsampling; this may be explicitly defined by one of the above properties for some input image items and/or the [=Sample Transform Derived Image Item=], and may be implicit for the other items (meaning no property defines chroma subsampling for these items);
```
wantehchang, line 416: Nit: Change "and/or" to "or".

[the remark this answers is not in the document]
leo-barnes: @wantehchang 
Are you ok with the current text? If so, let's merge so we can point the STF vote to it.

wantehchang (approved):

PR #316, opened 2025-08-29
workflow sync

podborski:
I would like to use the same process across different specs in STF group.

This PR is the initial attempt of a sync and RTP spec is used here as a reference.

It would be good to use the same bikeshed version (not the latest one) to avoid issues with bikeshed.

it would be nice to rename the master branch to main branch.


leo-barnes:
> * This PR makes sure that the same bikeshed version is used. This helps us to avoid issues with bikeshed versioning. I added requirements.txt that is currently using bikeshed 5.1.2

We should check what the version we are currently using is. Bikeshed made some fixes due to issues we filed a while back. We don't want to regress that if possible.

leo-barnes (approved):

